Issue #11, opened 2025-03-31
Build Automation - RFC


ssrowe:
## Overview
Implement automated build, packaging, and publishing capabilities for OIE to streamline development workflows and improve distribution.

### Priority 1: Core Build Automation
- Configure GitHub Actions to execute the current Ant-based build system
- Handle code signing
- Support seamless transition to a future build system (implementation details TBD)
- Create installable packages for all supported platforms (Windows, macOS, Linux)
- Implement automated verification testing as part of the build pipeline - not full application, unit or integration testing

### Priority 2: Artifact Publishing
- Publish library artifacts to GitHub Packages
  - Configure for future replication to additional repositories as needed
  - Include proper versioning and dependency management
  - Ensure appropriate documentation is generated and published

### Priority 3: Container Support
- Create and publish Docker images to GitHub Container Registry
  - Design for future replication to additional container registries as needed
  - Include multiple tags for version tracking (latest, specific versions)
  - Optimize container size and security

### Priority 3: Automated Testing & Releases (nice to have)
- Run automated tests as part of build process
- Automate the release process including Release, Release Notes

## Success Criteria
- Fully automated builds triggered by pull requests and daily builds
- Installable packages generated for each supported platform
- Libraries and Docker images published to GitHub and accessible by users
- Complete documentation for accessing and using published artifacts

## Supported Platforms
- Windows
  - 64-bit Installer (exe/msi)
  - 64-bit Zip
- Linux
  - Tar.gz
  - RPM
  - Installer (sh)
- macOS
  - Installer (dmg)
  - Tar.gz

Issue #3, opened 2025-03-29
Create tar/zip bundles


tonygermano:
Should include
- the compiled application
- the launcher from #2 
- ?

pacmano1:
This issue seems to be mixing topics.

There is create tar / zip bundles.  

There is "use the new server launcher script"

@tonygermano should this be open still?  Perhaps a new tissue for using the new launcher script.

PR #27, opened 2025-04-08
Add nightly builds

kpalang:
Add nightly builds of OIE

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - main
+  schedule:
+    - cron: '0 0 * * *' # Every 24 hours
 
 jobs:
   build:


jonbartels:
The scope of #3 is to make tarballs and zips

The scope of #11 is java support.

Neither requires nightly builds. 

Why not run builds on every merge to `main`. That way downloadable artifacts for the latest changes exist. Daily builds could add noise if we are not committing daily. Daily builds could miss changes if the team has a busy day.

tonygermano (commented):
I think I agree with @jonbartels that this probably isn't necessary since the automated builds were already running on every PR and push to main. I think the exception would be if the nightly tarballs could be published to a static link, so the url for the latest nightly never changed. Probably not worth the time to work on that at the moment since people can still grab the build with the latest changes from GH if they want it.

kpalang:
Good points. Closing this PR.

PR #44, opened 2025-04-19
DestinationSet should implement java.util.Set

rogin:
This was an old feature, requested by [several people](https://github.com/nextgenhealthcare/connect/issues/4414), that NextGen chose to ignore. Prior code review comments are [here](https://github.com/rogin/connect/pull/2). I compiled it with Java 8, and the test ran successfully. No further testing on my part.


tonygermano:
Thanks for submitting this! I always meant to give it a more thorough review and suggest a few things. Just a heads up, this probably will not be a top priority until after we get our first release out the door.

jonbartels (approved):

NicoPiel:
Can this be revisited now? @tonygermano

server/test/com/mirth/connect/userutil/DestinationSetTest.java
```diff
@@ -0,0 +1,376 @@
+
```
tonygermano: todo: new file needs SPDX header

server/test/com/mirth/connect/userutil/DestinationSetTest.java
```diff
@@ -0,0 +1,376 @@
...
+    @Test
+    public void testPopulatedSet() throws Exception {
+        DestinationSet ds = getNewTestDestinationSet();
+
+        assertFalse(ds.isEmpty());
+        assertEquals(3, ds.size());
+    }
+
+    @Test
+    public void testAdd() throws Exception {
```
tonygermano: I think `add` should throw `java.lang.UnsupportedOperationException`

server/test/com/mirth/connect/userutil/DestinationSetTest.java
```diff
@@ -0,0 +1,376 @@
...
+        //already in set
+        assertFalse(ds.add(1));
+        assertEquals(4, ds.size());
+
+        assertFalse(ds.add(null));
+        assertEquals(4, ds.size());
+    }
+
+    @Test
+    public void testAddAll() throws Exception {
```
tonygermano: I think `addAll` should throw `java.lang.UnsupportedOperationException`

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -7,162 +8,221 @@
...
+ import com.mirth.connect.userutil.ImmutableConnectorMessage;
+ 
+ /**
+  * Utility class used in the preprocessor or source filter/transformer to prevent the message from
+  * being sent to specific destinations.
+  */
+ public class DestinationSet implements Set<Integer> {
+ 
+     private Map<String, Integer> destinationIdMap = Collections.emptyMap();
+     private Set<Integer> metaDataIds;
```
tonygermano: I assume you are initializing `destinationIdMap` here in case it fails to set in the constructor. Can we initialize `metaDataIds` here as well instead of adding it to the catch block of the constructor? I think further discussion is needed on whether or not it's a good idea to change the behavior of DestinationSet not allowing additions. I think for the time being it can be initialized with `Collections.emptySet()`.

server/test/com/mirth/connect/userutil/DestinationSetTest.java
```diff
@@ -0,0 +1,376 @@
...
+    public void testToArrayGenericImproperlySized() throws Exception {
+        DestinationSet ds = getNewTestDestinationSet();
+
+        Integer[] resp = ds.toArray(new Integer[0]);
+        assertEquals(3, resp.length);
+        assertEquals(Integer.valueOf(1), resp[0]);
+        assertEquals(Integer.valueOf(2), resp[1]);
+        assertEquals(Integer.valueOf(3), resp[2]);
+    }
+}
```
tonygermano: nit: no newline at end of file

server/test/com/mirth/connect/userutil/DestinationSetTest.java
```diff
@@ -0,0 +1,376 @@
...
+    @Test
+    public void testToArray() throws Exception {
+        DestinationSet ds = getNewTestDestinationSet();
+
+        Object[] arr = ds.toArray();
+        assertEquals(3, arr.length);
+        assertEquals(Integer.valueOf(1), arr[0]);
+        assertEquals(Integer.valueOf(2), arr[1]);
+        assertEquals(Integer.valueOf(3), arr[2]);
+    }
```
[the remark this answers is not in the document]
NicoPiel: This is true. Tests that assume `toArray` to be deterministic could be flaky.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -7,162 +8,221 @@
...
+      *            The delegate ImmutableConnectorMessage object.
+      */
+     public DestinationSet(ImmutableConnectorMessage connectorMessage) {
+         try {
+             if (connectorMessage.getSourceMap().containsKey(Constants.DESTINATION_SET_KEY)) {
+                 this.destinationIdMap = connectorMessage.getDestinationIdMap();
+                 this.metaDataIds = (Set<Integer>) connectorMessage.getSourceMap().get(Constants.DESTINATION_SET_KEY);
+             }
+         } catch (Exception e) {
+             metaDataIds = new HashSet<>();
```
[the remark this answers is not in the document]
NicoPiel: This is also a good point and probable cause for future NPEs.

rogin:
I'm impressed by the copilot results. I'd like to resolve the findings this weekend.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -7,162 +8,221 @@
...
+ 
+     /**
+      * Stop a destination from being processed for this message.
+      * 
+      * @param metaDataIdOrConnectorName
+      *            An integer representing the metaDataId of a destination connector, or the actual
+      *            destination connector name.
+      * @return A boolean indicating whether at least one destination connector was actually removed
+      *         from processing for this message.
+      */
```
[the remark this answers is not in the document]
tonygermano: agree

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -56,15 +55,7 @@ public DestinationSet(ImmutableConnectorMessage connectorMessage) {
...
-        if (metaDataIds != null) {
-            Integer metaDataId = convertToMetaDataId(metaDataIdOrConnectorName);
-
-            if (metaDataId != null) {
-                return metaDataIds.remove(metaDataId);
-            }
-        }
-
-        return false;
+        return remove(Collections.singleton(metaDataIdOrConnectorName));
```
tonygermano, line 58: suggestion: annotate with `@Override`

This is a Set operation

---

suggestion: keep this method simple and don't wrap and delegate

If `metaDataIds` is made final and never null above, then the first null check can be removed.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -77,15 +68,15 @@ public boolean remove(Object metaDataIdOrConnectorName) {
...
-        boolean removed = false;
-
-        for (Object metaDataIdOrConnectorName : metaDataIdOrConnectorNames) {
-            if (remove(metaDataIdOrConnectorName)) {
-                removed = true;
-            }
-        }
-
-        return removed;
+        if(metaDataIdOrConnectorNames == null) { return false; }
```
tonygermano, line 71: nitpick: follow copilot recommendation to add space in `if (`

---

suggestion: move this implementation to Set operation `removeAll` and deprecate this method

This encourages the use of the Set operations `remove` for a single object and `removeAll` for collections. The overloaded `remove` has been a source of confusion in the past.

---

issue: should throw a NPE instead of return false when Collection is null

This matches the previous behavior and is specified by `Set.removeAll`

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -77,15 +68,15 @@ public boolean remove(Object metaDataIdOrConnectorName) {
...
+
+        return metaDataIdOrConnectorNames.stream()
+            .map(this::convertToMetaDataId)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(metaDataIds::remove)
+            .filter(Boolean::booleanValue)
+            .count() > 0;
```
tonygermano, line 79:
```suggestion
            .anyMatch(Boolean::booleanValue)
```

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -98,15 +89,7 @@ public boolean remove(Collection<Object> metaDataIdOrConnectorNames) {
...
-        if (metaDataIds != null) {
-            Integer metaDataId = convertToMetaDataId(metaDataIdOrConnectorName);
-
-            Set<Integer> set = (metaDataId != null) ? Collections.singleton(metaDataId) : Collections.emptySet();
-
-            return metaDataIds.retainAll(set);
-        }
-
-        return false;
+        return removeAllExcept(Collections.singleton(metaDataIdOrConnectorName));
```
tonygermano, line 92: suggestion: make this a call to Set operation `retainAll`

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -119,21 +102,15 @@ public boolean removeAllExcept(Object metaDataIdOrConnectorName) {
...
-                if (metaDataId != null) {
-                    set.add(metaDataId);
-                }
-            }
-
-            return metaDataIds.retainAll(set);
-        }
-
-        return false;
+        if(metaDataIdOrConnectorNames == null) { return false; }
```
tonygermano, line 105: nitpick: follow copilot recommendation to add space in `if (`

---

suggestion: move this implementation to Set operation `retainAll` and deprecate this method

This encourages the use of the Set operations and reduces confusion about two methods which do the same thing. I think `removeAllExcept(Object)` can remain since Set does not provide a `retainOnly` method for a single Object.

---

issue: should throw a NPE instead of return false when Collection is null

This matches the previous behavior and is specified by `Set.retainAll`

---

Suggestion: remove null check for metaDataIds

It should not be null if following other recommendations in this review

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -24,9 +22,9 @@
  * Utility class used in the preprocessor or source filter/transformer to prevent the message from
  * being sent to specific destinations.
  */
-public class DestinationSet {
+public class DestinationSet implements Set<Integer> {
 
-    private Map<String, Integer> destinationIdMap;
+    private Map<String, Integer> destinationIdMap = Collections.emptyMap();
     private Set<Integer> metaDataIds;
```
tonygermano, line 28: suggestion: make `destinationIdMap` and `metaDataIds` final and initialize both in constructor

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -43,6 +41,7 @@ public DestinationSet(ImmutableConnectorMessage connectorMessage) {
                 this.metaDataIds = (Set<Integer>) connectorMessage.getSourceMap().get(Constants.DESTINATION_SET_KEY);
             }
         } catch (Exception e) {
+            metaDataIds = new HashSet<>();
```
tonygermano, line 44: suggestion: when key is not present initialize to `Collections.emptySet()`

Consider making fields final and modifying the constructor like this.

```java
public DestinationSet(ImmutableConnectorMessage connectorMessage) {
    Map<String, Integer> tempMap = Collections.emptyMap();
    Set<Integer> tempSet = Collections.emptySet();

    if (connectorMessage != null) {
        try {
            Object rawValue = connectorMessage.getSourceMap().get(Constants.DESTINATION_SET_KEY);
            
            if (rawValue instanceof Set<?> set) {
                tempMap = connectorMessage.getDestinationIdMap();
                
                @SuppressWarnings("unchecked")
                Set<Integer> castedSet = (Set<Integer>) set;
                tempSet = castedSet;
            }
        } catch (Exception e) {
            // consider logging an error, but logging is not currently set up for this class
            // logger.error("Failed to extract destination metadata from connector message", e);
        }
    }

    this.destinationIdMap = tempMap;
    this.metaDataIds = tempSet;
}
```

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
-        }
-
-        return false;
+        int origSize = size();
+        clear();
+        return origSize > 0;
     }
 
-    private Integer convertToMetaDataId(Object metaDataIdOrConnectorName) {
+    private Optional<Integer> convertToMetaDataId(Object metaDataIdOrConnectorName) {
```
tonygermano, line 129: suggestion: keep return value for this method as it is

Since this method is private, it's not really necessary to wrap and unwrap with an Optional in the same class. The nulls are cleaner.

In your streams you could do
```diff
+import java.util.Objects;
-            .filter(Optional::isPresent)
-            .map(Optional::get)
+            .filter(Objects::nonNull)
```

I don't think it is necessary to replace the multiple returns either, but it is a good change to remove the destinationIdMap null check if we always initialize it to be non-null.

It could be helpful to add a javadoc for this method to describe its behavior.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+    @Override
+    public boolean contains(Object metaDataIdOrConnectorName) {
+        Optional<Integer> m = convertToMetaDataId(metaDataIdOrConnectorName);
+
+        return m.isPresent() && metaDataIds.contains(m.get());
+    }
+
+    @Override
+    public Iterator<Integer> iterator() {
+        return Collections.unmodifiableSet(metaDataIds).iterator();
```
tonygermano, line 164: issue: iterator should be modifiable

The only method on the iterator which allows modifying the Set is `remove` which is an allowed operation.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+        return metaDataIds.toArray();
+    }
+
+    @Override
+    public <T> T[] toArray(T[] a) {
+        return metaDataIds.toArray(a);
+    }
+
+    @Override
+    public boolean add(Integer metaDataId) {
```
tonygermano, line 178: issue: `add` should throw UnsupportedOperationException

DestinationSets do not support adding to the Set

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+    public boolean containsAll(Collection<?> metaDataIdOrConnectorNames) {
+        if(metaDataIdOrConnectorNames == null) { return false; }
+
+        return metaDataIdOrConnectorNames.stream()
+        .map(this::contains)
+        .allMatch(Boolean::booleanValue);
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends Integer> metaDataIdOrConnectorNames) {
```
tonygermano, line 192: issue: `addAll` should throw UnsupportedOperationException

DestinationSets do not support adding to the Set

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+                    changed = true;
+                }
+            }
+        }
+
+        return changed;
+    }
+
+    @Override
+    public boolean retainAll(Collection<?> metaDataIdOrConnectorNames) {
```
tonygermano, line 209: suggestion: deprecate `removeAllExcept(Collection<Object>)` in favor of this method

Also mentioned above. The Set operations should be the primary methods to reduce confusion and we can eventually remove the old methods.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+        return removeAllExcept((Collection<Object>)metaDataIdOrConnectorNames);
+    }
+
+    @Override
+    public boolean removeAll(Collection<?> metaDataIdOrConnectorNames) {
```
tonygermano, line 214: suggestion: deprecate `remove(Collection<Object>)` in favor of this method

Also mentioned above. The Set operations should be the primary methods to reduce confusion and we can eventually remove the old methods.

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+    @Override
+    public boolean isEmpty() {
+        return metaDataIds.isEmpty();
+    }
+
+    @Override
+    public boolean contains(Object metaDataIdOrConnectorName) {
+        Optional<Integer> m = convertToMetaDataId(metaDataIdOrConnectorName);
+
+        return m.isPresent() && metaDataIds.contains(m.get());
```
tonygermano, line 159: if `convertToMetaDataId` is changed back to returning nulls and `metaDataIds` is always non-null as suggested elsewhere, I think this can be simplified to just

```java
return metaDataIds.contains(convertToMetaDataId(metaDataIdOrConnectorName));
```

server/src/com/mirth/connect/server/userutil/DestinationSet.java
```diff
@@ -144,25 +121,102 @@ public boolean removeAllExcept(Collection<Object> metaDataIdOrConnectorNames) {
...
+    }
+
+    @Override
+    public boolean add(Integer metaDataId) {
+        return metaDataId != null && metaDataIds.add(metaDataId);
+    }
+
+    @Override
+    public boolean containsAll(Collection<?> metaDataIdOrConnectorNames) {
+        if(metaDataIdOrConnectorNames == null) { return false; }
```
tonygermano, line 184: nitpick: follow copilot recommendation to add space in `if (`

---

issue: `Set.containsAll` should throw a NPE instead of returning false when Collection is null

PR #45, opened 2025-04-19
Maintain consistent element order in exports

rogin:
Fixes an [old and common complaint](https://github.com/nextgenhealthcare/connect/issues/5004). Previous test results [here](https://github.com/nextgenhealthcare/connect/pull/5791). I've only verified that it built successfully using Java 8.


jonbartels (approved):

gibson9583 (approved):

kpalang (dismissed):

donkey/src/main/java/com/mirth/connect/donkey/model/message/MapContent.java
```diff
@@ -9,11 +9,11 @@
 
 package com.mirth.connect.donkey.model.message;
 
-import java.util.HashMap;
+import java.util.TreeMap;
```
tonygermano, line 12: Since a TreeMap is sorted and a HashMap is not, I'm guessing there could be considerable changes compared to a prior export the first time an export is done with the new code?

server/src/com/mirth/connect/model/converters/MapContentConverter.java
```diff
@@ -53,7 +53,7 @@ public void marshal(Object value, HierarchicalStreamWriter writer, MarshallingCo
             try {
                 DonkeyElement mapElement = new DonkeyElement(serializedMap);
                 mapElement.setNodeName("content");
-                mapElement.setAttribute("class", "map");
+                mapElement.setAttribute("class", "tree-map");
```
tonygermano: what exactly does this impact?

donkey/src/main/java/com/mirth/connect/donkey/model/message/MapContent.java
rogin: Correct, then subsequent exports will benefit.

server/src/com/mirth/connect/model/converters/MapContentConverter.java
rogin: I don't recall the original reason, and removing it changes nothing in the exports I performed today, so it can be omitted.

donkey/src/main/java/com/mirth/connect/donkey/model/message/MapContent.java
tonygermano: Is there a purpose for having MapContent as part of this PR? I think it only shows up in a message export, and not part of the server configuration?

rogin: I'll revert it and verify.

jonbartels:
> Can you describe your testing? I'm sure this does work when creating and exporting new code templates and libraries. Do we know for certain that when xstream creates the objects from xml that it isn't recreating them as HashSets?

PR #47, opened 2025-04-19
add two vars to the alert template variable list

rogin:
Adds "serverName" and "environmentName" to the alert template variable list.
Requires Java 9+.

Fixes NextGen request [#6261](https://github.com/nextgenhealthcare/connect/issues/6261).

Previous PR is [#6262](https://github.com/nextgenhealthcare/connect/pull/6262).

Slight tangent: I deleted `KeyEncryptorTest` to get over / ignore its java 9+ error that I don't care to investigate as I believe it's unrelated to this code change: `error: package com.sun.crypto.provider is not visible`.

**Assuming we have documentation, it should be updated as two additional variables will display in the Alert Template variable list.**

Test procedure:
1. Compile this patch with Java 9+ and start server
2. Login and get past initial account setup
3. Set Server Name and Environment Name under Settings tab
4. Import and deploy the two attached channels
5. Import the attached alert (remove its .txt extension after downloading)
6. Open the imported alert and validate two additional options are present on the right-hand side: "serverName" and "environmentName"
7. Validate the two additional options can be copied into the template text area
8. Send any message to channel "throws error"
9. Validate the resulting text in channel "Receiver" contains the Server Name and Environment Name as you defined in the Settings tab
10. Clear the Server Name and Environment Name under Settings tab, then re-run a message to ensure the alert template's values are replaced with empty strings

[throws.error.xml.txt](https://github.com/user-attachments/files/19818413/throws.error.xml.txt)
[Receiver.xml.txt](https://github.com/user-attachments/files/19818412/Receiver.xml.txt)
[Foo.alert.xml.txt](https://github.com/user-attachments/files/19818411/Foo.alert.xml.txt)


jonbartels (dismissed):

rogin:
You can tell how often I use git for these situations. Need help getting just my commit signed and pushed.

tonygermano (changes requested):
I've rebased this and resolved the merge conflicts since #258 was merged. It works, but I'm reluctant to approve at this time.

The currently used `ConfigurationController.getInstance().getServerName()` executes quickly because the server name is cached by the DefaultConfigurationController.

`ConfigurationController.getInstance().getServerSettings()`, however, is a much more expensive call, as it makes multiple calls to the database each time.

The expensive call has the benefit of returning up-to-date information, even if the settings were changed by a different server instance connected to the same database. I wonder if we should add a separate `getCachedServerSettings` method on `ConfigurationController` that this would call instead?

I have not done any benchmarking. I'll defer to others if consensus is that these extra db calls should not be considered a problem.

PR #48, opened 2025-04-19
Update WordUtils from deprecated version

rogin:
Usage of `WordUtils` was showing as deprecated.

Code was using `WordUtils` from common-lang3 jar where it [became deprecated at v3.6](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/text/WordUtils.html) (we're using v3.13). This change uses `WordUtils` from commons-text as recommended. As third party libs aren't shared across subprojects, I had to duplicate the commons-text jar under donkey/lib/commons/.


jonbartels (approved):
LGTM

kpalang (approved):

PR #60, opened 2025-04-24
Add tiered vmoptions to local builds

tonygermano:
Part of #2 

In the nextgen releases, the vmoptions files were only included with the installation packages, but not in the source tree. This PR adds a tiered vmoptions configuration which gets included at build time.


kayyagari (approved):

server/build.properties
jonbartels, line 1: Whats with the line endings?

jonbartels (changes requested):

server/build.properties
tonygermano: Typical best practice is to keep the same line endings that already exist in the file rather than mixing line endings or rewriting the entire file. This file already had Windows line endings, so I kept it that way.

pacmano1 (approved):
Chatted with @tonygermano on a few items on how the installer handles this.  Otherwise no comments.

jonbartels (approved):

PR #61, opened 2025-04-28
Create launcher to parse vmoptions and start server

### LINKED ISSUE #2: Create server launcher to replace mcserver
mcserver was created by install4j, and the configuration was not open source. I don't believe that a replacement for mcservice is required, as each platform has a preferred way to run a service, and they should all work with a simple launcher.

needs:
- a binary or script that can run with a double-click (source code must be open)
- must process a .vmoptions file as described [here](https://www.ej-technologies.com/resources/install4j/help/doc/concepts/vmParameters.html) or something similar in a different documented format.
  - in addition to directly using options as arguments to the java executable, it supports special values for:
    - replacing, appending to, or prepending to the classpath
    - including other vmoptions files, with environment variable substitution in the path
- ensure that the classpath is built in the correct order. Certain project libraries reimplement classes from other libraries and must come earlier on the classpath.
- able to locate a jvm with which to launch the server

nice-to-haves:
- easy (and well documented) way to configure it to launch using a specified jvm. Shouldn't require a system-wide jvm to be installed.
- detect java version and automatically include known required options for java 9+

Platform support (check as completed):
- [ ] Windows
- [ ] macOS
- [ ] Linux (+docker)

tonygermano:
closes #2

See README for details

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,6 +22,11 @@ jobs:
           java-package: 'jdk+fx'
           distribution: 'zulu'
 
+      - name: Install clojure tools
+        uses: DeLaGuardo/setup-clojure@13.2
+        with:
+          lein: 2.11.2
+
       - name: Build OIE (signed)
         if: github.ref == 'refs/heads/main'
         working-directory: server
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -2,3 +2,4 @@
 # Add key=value pairs of SDKs to use below
 java=8.0.442.fx-zulu
 ant=1.10.14
+leiningen=2.11.2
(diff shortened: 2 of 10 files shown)


tonygermano:
This has worked well in my testing (only on linux,) but it wasn't really what I intended for it to be. I figured I'd get it out for review so we can use it in the short term even if we end up replacing it with something else down the road. My plan was for it to be compiled to native using graalvm native-image, but I was having issues with that. It took a very long time to compile (around 6 minutes, compared to a few seconds for the jar version) produced a binary over 50MB in size, and then it was throwing exceptions during the shutdown hook.

So, I decided for now it would just run using java (can be the same java version as what ultimately launches the server or a different one.)

jonbartels (approved):

kayyagari (approved):

Issue #65, opened 2025-04-30
[BUG] External web links not opening on some Linux systems


tonygermano:
Clicking on any of the links in the Administrator that launch a web browser (such as the options to view the User API or Client API) fail with an error that the web browser could not be found.

Issue was noticed when running the Administrator on a Chromebook in the standard ChromeOS Linux container. No additional browsers are installed besides Chrome as part of ChromeOS. Other Linux apps are able to launch links in the host OS browser.

PR #66, opened 2025-04-30
Upgrade BareBonesBrowserLauncher; improve linux support

tonygermano:
Upgraded from 1.5 to latest version 3.2. Fixed deprecated error. Added xdg-open as first choice on linux systems.

fixes OpenIntegrationEngine/engine#65

### DIFF
--- a/client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
+++ b/client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
@@ -1,47 +1,73 @@
-// ///////////////////////////////////////////////////////
-// Bare Bones Browser Launch //
-// Version 1.5 //
-// December 10, 2005 //
-// Supports: Mac OS X, GNU/Linux, Unix, Windows XP //
-// Example Usage: //
-// String url = "http://www.centerkey.com/"; //
-// BareBonesBrowserLaunch.openURL(url); //
-// Public Domain Software -- Free to Use as You Like //
-// ///////////////////////////////////////////////////////
+// SPDX-License-Identifier: MPL-2.0
+// SPDX-FileCopyrightText: 2025 Tony Germano <tony@germano.name>
+// SPDX-Contributor: Original Author Dem Pilafian (Original work licensed under WTFPL)
+
 package com.mirth.connect.client.ui;
 
-import java.lang.reflect.Method;
+import java.util.Arrays;
 
+/**
+ * Utility class to open a web page from a Swing application
+ * in the user's default browser.
+ * <p>
+ * Supports: Mac OS X, Linux, Unix, Windows
+ * <p>
+ * Example usage:<br>
+ * <code>&nbsp;&nbsp;
+ *    String url = "https://dna-engine.org/";<br>&nbsp;&nbsp;
+ *    BareBonesBrowserLaunch.openURL(url);</code>
+ * <p>
+ * Latest Version: <a href=https://centerkey.com/java/browser>
+ * https://centerkey.com/java/browser</a>
+ * <p>
+ * Published: October 24, 2010
+ * Modified: 2025
+ * <p>
+ *
+ * @author Dem Pilafian
+ * @version 3.2
+ */
 public class BareBonesBrowserLaunch {
 
+    static final String[] browsers = { "xdg-open", "x-www-browser", "google-chrome",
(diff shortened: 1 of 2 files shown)


kayyagari (approved):

client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
```diff
@@ -1,47 +1,73 @@
...
-                    throw new Exception("Could not find web browser");
-                } else {
-                    Runtime.getRuntime().exec(new String[] { browser, url });
+        try { // attempt to use Desktop library from JDK 1.6+
+            Class<?> d = Class.forName("java.awt.Desktop");
+            d.getDeclaredMethod("browse",
+                    new Class<?>[] { java.net.URI.class }).invoke(
+                            d.getDeclaredMethod("getDesktop").invoke(null),
+                            new Object[] { java.net.URI.create(url) });
+        } catch (Exception ignore) { // library not available or failed
```
jonbartels, line 49: Is this using exceptions as flow control?

client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
```diff
@@ -1,47 +1,73 @@
...
+ * Published: October 24, 2010
+ * Modified: 2025
+ * <p>
+ *
+ * @author Dem Pilafian
+ * @version 3.2
+ */
 public class BareBonesBrowserLaunch {
 
+    static final String[] browsers = { "xdg-open", "x-www-browser", "google-chrome",
```
jonbartels, line 32: Put a comment saying where this list is sourced from and when it might have to change.

client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
```diff
@@ -1,47 +1,73 @@
...
-                    if (Runtime.getRuntime().exec(new String[] { "which",
-                            browsers[count] }).waitFor() == 0) {
-                        browser = browsers[count];
-                    }
-                }
-                if (browser == null) {
-                    throw new Exception("Could not find web browser");
-                } else {
-                    Runtime.getRuntime().exec(new String[] { browser, url });
+        try { // attempt to use Desktop library from JDK 1.6+
```
jonbartels, line 43: Is this library still present in modern JDKs?

client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
```diff
@@ -1,47 +1,73 @@
...
+                else { // assume Unix or Linux
+                    String browser = null;
+                    for (String b : browsers)
+                        if (browser == null
+                                && Runtime.getRuntime().exec(new String[] { "which", b }).getInputStream().read() != -1)
+                            Runtime.getRuntime().exec(new String[] { browser = b, url });
+                    if (browser == null)
+                        throw new Exception(Arrays.toString(browsers));
                 }
+            } catch (Exception e) {
```
jonbartels, line 68: Are pokemon exceptions appropriate here?

manager/src/com/mirth/connect/manager/BareBonesBrowserLaunch.java
jonbartels, line 1: Why is this file duplicated between client and manager?

jonbartels (changes requested):

jonbartels (approved):

manager/src/com/mirth/connect/manager/BareBonesBrowserLaunch.java
tonygermano: This is because the method of alerting the user in the final catch is different between the two files.

client/src/com/mirth/connect/client/ui/BareBonesBrowserLaunch.java
tonygermano: Yes, [it is still present](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/Desktop.html), but the older version of the software that was present in mirth was written for java 1.5, so the class did not exist at the time. However, the Linux support is not very good so the fallback mechanism is still needed. My searching tells me that it only supports gnome desktops, and on my non-gnome system it threw an UnsupportedOperationException when I tried to call the browse method.

PR #68, opened 2025-04-30
Issue 6 first login required fields

### LINKED ISSUE #6: Only username and password should be required fields for first time login

jonbartels:
Resolves #6 

First login and user edit only require username and password

user-edit

first-login

### DIFF
--- a/client/src/com/mirth/connect/client/ui/FirstLoginDialog.java
+++ b/client/src/com/mirth/connect/client/ui/FirstLoginDialog.java
@@ -47,7 +47,7 @@ public class FirstLoginDialog extends javax.swing.JDialog implements UserDialogI
         finishButton.setEnabled(false);
 
         userEditPanel.setUser(this, currentUser);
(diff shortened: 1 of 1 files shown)


gibson9583 (approved):

kpalang (approved):

tonygermano (changes requested):
I'd like to take a look at rolling back this class to what it looked like in 4.0 before they started making the changes that required all of the additional fields. There may be a few things we need to keep, because I think a few extra fields were added, but I know a lot of GUI elements were added or hidden which we don't need or want. We likely don't need the changes that pre-populate the form with values saved in the local java user profile if we aren't requiring the fields, anymore. I think comparing what is there currently to a known desired state is probably easier than trying to slowly undo things piece by piece and likely overlooking something.

My apologies for not including this in the detail of the issue.

tonygermano (approved):
I'll open a new issue for the rollback.

Issue #24, opened 2025-04-08
Implement sample notifications server


kpalang:
Administrator has the ability to [query](https://github.com/OpenIntegrationEngine/engine/blob/37968e15b32b60e1efb33d66aa6762ae65d9cd95/server/src/com/mirth/connect/client/core/ConnectServiceUtil.java#L90) and display dynamic notifications. We should write a reference implementation.

jonbartels:
This is a good PR for anyone with any programming skills. It should be a very simple REST API.

Key acceptance criteria:
* Messages should be able to be created through normal release or PR processes
* Reading from existing processes will be helpful - example can the code read from an "Announcements.md" in the repo or GitHub releases?
* Target URLs should be configurable with other branding configs, see https://github.com/OpenIntegrationEngine/engine/pull/23 . That way vendors or resellers or other takers can set up their own notifications

The implementation could be EITHER a server or serverless thing OR changes to the Mirth code to just read from GitHub or openintegrationengine.org

leovander:
Repos have a few atom feeds to read from, potentially for this issue using this feed could function as the source of truth for `https://github.com/OpenIntegrationEngine/engine/releases.atom`.

Here is a sample entry returned from above:
```xml
<entry>
    <id>tag:github.com,2008:Repository/956275223/4.5.2</id>
    <updated>2024-09-10T06:12:21-07:00</updated>
    <link rel="alternate" type="text/html" href="https://github.com/OpenIntegrationEngine/engine/releases/tag/4.5.2"/>
    <title>4.5.2</title>
    <content type="html">&lt;p&gt;Tagged &lt;a class=&quot;commit-link&quot; data-hovercard-type=&quot;commit&quot; data-hovercard-url=&quot;https://github.com/OpenIntegrationEngine/engine/commit/6ce3a9f0e3d84841f0b1e07c2808cf0bdb3d0a78/hovercard&quot; href=&quot;https://github.com/OpenIntegrationEngine/engine/commit/6ce3a9f0e3d84841f0b1e07c2808cf0bdb3d0a78&quot;&gt;&lt;tt&gt;6ce3a9f&lt;/tt&gt;&lt;/a&gt;&lt;/p&gt;</content>
    <author>
      <name>tonygermano</name>
    </author>
    <media:thumbnail height="30" width="30" url="https://avatars.githubusercontent.com/u/39179051?s=60&amp;v=4"/>
  </entry>
```

[...]

PR #79, opened 2025-05-05
Notifications

kelaompachai:
Update client notifications to pull from github Release API
- Added library java-semver-0.10.2.jar for version parsing and comparison.
- Refactored to use streams and added tests
- No longer calls a self-hosted service, so active extensions are no
  longer provided as part of the request.


jonbartels (changes requested):
Add `.vscode` and `.project` to `.gitignore`

Thank you for the PR!

rogin:
I've cleaned up the code and added tests. I've drafted a PR to @kelaompachai 's main.

kelaompachai:
I rebased to get rid of the out-of-scope commits that had to do with .project, .vscode, .gitignore.

server/test/com/mirth/connect/client/core/ConnectServiceUtilTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.apache.http.entity.InputStreamEntity;
+import org.apache.http.message.BasicHeader;
+import org.junit.Test;
+
+public class ConnectServiceUtilTest {
+    private static final String FILE_POPULATED = "notifications-populated.json";
+    private static final String FILE_EMPTY = "notifications-empty.json";
+
+    private InputStreamEntity createEntityFromFile(String filename) {
+        var is = getClass().getClassLoader().getResourceAsStream(filename);
```
jonbartels: Don't use `var` - We are currently compiling under Java 8 to mirror MC 4.5.2 (see: https://github.com/OpenIntegrationEngine/engine/blob/main/.github/workflows/build.yaml#L21)

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,62 +97,181 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
 
-                for (JsonNode childNode : rootNode) {
-                    Notification notification = new Notification();
-                    notification.setId(childNode.get("id").asInt());
-                    notification.setName(childNode.get("name").asText());
-                    notification.setDate(childNode.get("date").asText());
-                    notification.setContent(childNode.get("content").asText());
-                    allNotifications.add(notification);
-                }
+                var newerOnly = filterForNewerVersions(toJsonStream(responseEntity), mirthVersion);
```
jonbartels: Don't use var - We are currently compiling under Java 8 to mirror MC 4.5.2 (see: https://github.com/OpenIntegrationEngine/engine/blob/main/.github/workflows/build.yaml#L21)

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,62 +97,181 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
+
+    /**
+     * Compare the current version with another. If current is greater than or equal, this returns false.
+     * 
+     * @param currentVersion
+     * @param anotherVersion
+     * @return true if the current version is less than than the other
+     */
+    protected static boolean isCurrentOlderThan (int[] currentVersion, String anotherVersion) {
+        return Arrays.compare(currentVersion, toVersionArray(anotherVersion)) < 0;
```
jonbartels: Arrays.compare doesn't exist in Java 8

We are currently compiling under Java 8 to mirror MC 4.5.2 (see: https://github.com/OpenIntegrationEngine/engine/blob/main/.github/workflows/build.yaml#L21)

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,62 +97,181 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
+    /**
+     * Filter the stream to only new versions, then create a notification for each.
+     * 
+     * @param nodes
+     * @param currentVersion
+     * @return a non-null List
+     */
+    protected static Stream<JsonNode> filterForNewerVersions(Stream<JsonNode> nodes, String currentVersion) {
+        int[] curVersion = toVersionArray(currentVersion);
+        return nodes.takeWhile(node -> isCurrentOlderThan(curVersion, node.get("tag_name").asText()));
```
jonbartels: Stream.takeWhile wasn't introduced until Java 9, we're still compiling on Java 8

This alternative from AI seems reasonable - https://chatgpt.com/share/682762dc-fb40-800e-9de1-a56febb7ba26

jonbartels (changes requested):
Several Java 8 compatibility issues.

rogin:
I'm out of gas trying to merge correctly. I have java 8 compatible code sitting on [this commit](https://github.com/OpenIntegrationEngine/engine/commit/958004c431b75bdc860bb8557cdaae5d3dd8263e). I'd appreciate help, otherwise it's dead on the vine.

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
kelaompachai: I think we should be able to just use filter there. I think takeWhile is just a performance improvement that breaks out/returns earlier by assumimg newer versions come earlier in the array sent back from the github releases api.

kelaompachai:
I got rogin's java 8 compatible code working on my machine and pushed it up into this pr.

kelaompachai:
I did have to manually edit the server/.classpath file to get it to work.

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,62 +99,187 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
+    }
+
+    /**
+     * Create a request config with appropriate network timeouts.
+     * 
+     * @return
+     */
+    private static RequestConfig createRequestConfig() {
+        return RequestConfig.custom().setConnectTimeout(TIMEOUT).setConnectionRequestTimeout(TIMEOUT).setSocketTimeout(TIMEOUT).build();
+    }
```
tonygermano: I haven't had time to test with the PR, but using curl I can add `-H "Accept: application/vnd.github.html+json"` to return a `body_html` property as rendered html instead of a `body` property using markdown. This may allow the details of the release to be easily inserted into the notification.

I'm assuming this is the function where that header would be added.

tonygermano: And if it ends up looking terrible, I found this https://github.com/sindresorhus/github-markdown-css

kelaompachai: Ok, I'm working on it.

kelaompachai: This commit takes the html from the github api and directly inserts it into the notification content. Is that a security concern?

Some release pages would look fine, others would look bad. Nextgen's for example looks bad because there are images that the oie engine fails to fetch. It would be up to whoever is in control of the releases page to ensure that it would look good.

jonbartels:
I did some testing on Saturday.

----

> Built from keloampachi main branch. Did not mess with versions or any defaults
> No notifications shown; no errors

**Correction needed** - there are no notifications to show, then do not show the notification popup.

image

----

----

> Manually forcing the version to 3.11 shows 

Links are clickable

***This test case passed*** this is the main workflow. GJ!

***Improvement requested***:
I think the PR should show more data like we see from `https://github.com/nextgenhealthcare/connect/releases/tag/4.4.2`
> Mirth Connect 4.4.2 is a patch release that includes bug fixes and an improvement to exporting messages.
Then the whats new, upgrade, and release links.

Is that data included in the payload returned from git or would it require a second fetch?

image

----

kelaompachai:
The most recent commit should address the requested more data improvement.

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
kelaompachai: Now the images are fetching for me. Still doesn't look great though.

tonygermano: would you be able to share a screenshot on this PR of the current state?

kelaompachai: Screenshot 2025-05-19 at 10 25 24 PM

kelaompachai: Using nextgen's releases since we don't have any yet.

The images are inconsistent, and in my opinion, there is inessential information and having the left-to-right scrollbar is less than ideal.

kelaompachai: Screenshot 2025-05-19 at 10 34 26 PM

kelaompachai: But if the string is cut right before the Downloads header, it looks quite good.

jonbartels:
I pushed changes to not show the dialog when there are no notifications

jonbartels (dismissed):

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,110 +99,178 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
 
-            post.setURI(URI.create(URL_CONNECT_SERVER + URL_NOTIFICATION_SERVLET));
-            post.setEntity(new UrlEncodedFormEntity(Arrays.asList(params), Charset.forName("UTF-8")));
+    /**
+     * Split a given version string into an int[].
+     * 
+     * @param version
+     * @return
+     */
+    protected static int[] toVersionArray(String version) {
```
rogin: This is no longer used and should be deleted.

rogin: And all tests which call it are no longer necessary.

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
```diff
@@ -87,110 +99,178 @@ public static void registerUser(String serverId, String mirthVersion, User user,
...
+        return notification;
+    }
+
+    /**
+     * Create the HTML content for a notification.
+     * 
+     * @param node
+     * @return an HTML String
+     */
+    protected static String toNotificationContent (JsonNode node) {
```
rogin: As it's a one-liner now and called once, consider moving it to lines 252-253.

server/test/com/mirth/connect/client/core/ConnectServiceUtilTest.java
```diff
@@ -0,0 +1,206 @@
...
+
+        assertTrue("A node was expected to act upon", firstNode.isPresent());
+        
+        JsonNode orig = firstNode.get();
+
+        Notification created = ConnectServiceUtil.toNotification(orig);
+        assertEquals(orig.get("id").asInt(), (int)created.getId());
+        assertEquals(orig.get("name").asText(), created.getName());
+        assertEquals(orig.get("published_at").asText(), created.getDate());
+        assertNotNull(created.getContent());
```
rogin: Could now be _assertEquals(orig.get("body_html").asText(), created.getContent());_

jonbartels (dismissed):

rogin:
Brainstorming idea on a [commit](https://github.com/OpenIntegrationEngine/engine/commit/55f51533b455b8473c2b953195dfea7eb093bfeb).

Maybe it's over designing, but I noticed the branding branch and refactored the release URL to `BrandingConstants` which they created. From there, it makes sense for `getNotifications` to be an interface so others can read versions from a file or other source. So I wrote a default implementation that will process a Github project URL declared at `BrandingConstants.RELEASES_URL`, and those that read from a file can specify a different class that implements the interface. Since changing the "Accept" header didn't return the crisp data we were hoping for, I omitted it.

server/test/com/mirth/connect/client/core/ConnectServiceUtilTest.java
kelaompachai: That test is failing when I build and I can't figure out why.

kelaompachai: Ok, I figured it out.

tonygermano:
> Brainstorming idea on a [commit](https://github.com/OpenIntegrationEngine/engine/commit/55f51533b455b8473c2b953195dfea7eb093bfeb).
> 
> Maybe it's over designing, but I noticed the branding branch and refactored the release URL to `BrandingConstants` which they created. From there, it makes sense for `getNotifications` to be an interface so others can read versions from a file or other source. So I wrote a default implementation that will process a Github project URL declared at `BrandingConstants.RELEASES_URL`, and those that read from a file can specify a different class that implements the interface. Since changing the "Accept" header didn't return the crisp data we were hoping for, I omitted it.

I figured it was out of scope for this PR, but at some point I'd like to make a plugin interface for notification suppliers, and this could become a reference implementation. I know there have been requests in the past for people to be able to create their own notifications to send to their users and may prefer a notification source other than github releases. I didn't want those ideas to slow progress on this PR, which I think does a fine job of filling a short-term need.

jonbartels (approved):

kpalang:
The implementation looks quite impressive and massive kudos to @kelaompachai for building it! But there's one but I'm having a hard time overlooking.
The current approach is very GitHub API-centric. So-much-so that any other server would require substantial rewrite. I feel bad bringing this out at such a late stage of the PR, but there were mentions under #24 of serving notifications via RSS or Atom, which in my opinion would be a more flexible approach to delivering notifications. This way any downstreams would simply have to provide a feed url rather than rewrite the fetching and rendering logic.

Another thing I see little value in is having the notifications version-specific. In my opinion we don't really have that much notifications and info specific to a single version that this more useful than it is complex.

kayyagari (approved):

gibson9583 (approved):
LGTM

server/src/com/mirth/connect/client/core/ConnectServiceUtil.java
tonygermano: For now we will leave it as it. We can make adjustments if our project's releases cause display issues.

Issue #78, opened 2025-05-02
[SECURITY] Update Postgres JDBC driver (https://nvd.nist.gov/vuln/detail/CVE-2024-1597)


pacmano1:
**Describe the security issue**
Upgrade Postgres JDBC driver.

**Vulnerability Location**
Third Party
(https://nvd.nist.gov/vuln/detail/CVE-2024-1597)

**Environment (please complete the following information if it is applicable to the issue)**
All

**Suggested remediation**
postgresql-42.6.0.jar -> postgresql-42.6.1.jar

**Additional context**
```
└─$ find . -name "postgres*jar" -print                                                                                                               1 ⨯
./engine/donkey/setup/lib/database/postgresql-42.6.0.jar
./engine/donkey/lib/database/postgresql-42.6.0.jar
./engine/server/setup/server-lib/database/postgresql-42.6.0.jar
./engine/server/lib/database/postgresql-42.6.0.jar
./engine/simplesender/lib/postgresql-8.1-405.jdbc3.jar
```

PR #85, opened 2025-05-10
Issue #78 update PG drivers

rogin:
Issue #78

Pulled latest v42.6.x from [here](https://repo1.maven.org/maven2/org/postgresql/postgresql/42.6.2/). SHA1 hash verified. Sonatype showed it having zero vulnerabilities.

Could not find other PG jar at version "-405". Pulled latest "-408" version [here](https://repo1.maven.org/maven2/postgresql/postgresql/8.1-408.jdbc3/). 2007 timestamp. SHA1 hash verified.

Code compiled w/exception of KeyEncryptorTest, which I deleted. All failed tests ignored.

Ignore the tool listing two renames - two files were deleted, and two files were added.


tonygermano:
I think I would not bother updating the driver for simplesender. I didn't even realize what that code did until I looked at it. It hasn't been touched in 13 years. That library was added 18 years ago. My best guess is that it was part of some undocumented test process. I don't believe it was included in the mirth distributions, and it doesn't look like it even gets built as part of the normal build process.

We probably need to evaluate to see if we want to get rid of it, and if we decide to keep it with the existing postgres functionality, it should be using a modern driver.

jonbartels (approved):

jonbartels (changes requested):
Do not update the JARS in simplesender. They are not packaged in the OIE server or client.

jonbartels:
@tonygermano - I agree with you. I filed https://github.com/OpenIntegrationEngine/engine/issues/88

jonbartels (approved):

pacmano1 (approved):

pacmano1:
Ignore me asking CoPilot to review please.

Issue #87, opened 2025-05-14
[IDEA] Add the ability to format javascript code


sbraconnier:
**Is your feature request related to a problem? Please describe.**
When editing javascript code, it's not possible to format the code using a quick action (not that I'm aware of). It would be useful to be able to format the code using a context menu and/or a keyboard shortcut.

**Describe your use case**
na.

**Describe the solution you'd like**
The provided [js-beautify](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/util/beautify-1.6.8.js) library could be used to create an new [action](https://github.com/OpenIntegrationEngine/engine/tree/main/client/src/com/mirth/connect/client/ui/components/rsta/actions) for the rsta that will allow the user to format the javascript code.

**Describe alternatives you've considered**
na.

**Additional context**
I would be happy to contribute for this one.

tonygermano:
I think that's a great idea. Do you know if js-beautify works with rhino code containing all of the odd e4x syntax and operators? That seems to be the issue with most beautifiers and js written for mirth.

tonygermano:
Looks viable. I just tried the following at https://js-beautify.com

```javascript
const val = 1

default         xml namespace = 'uri://test'
  var x = <xml>
      <a val={val}/>
 </xml>
  
 for each (var a in x   ..a) { var b = a.*::b
               }
```

There was a setting at the top I had to change for e4x/jsx support or it really messed up the xml literal. It still didn't change the indents for the xml,  but otherwise it seemed to work fine and cleaned it up to
```javascript
const val = 1

default xml namespace = 'uri://test'
var x = <xml>
      <a val={val}/>
 </xml>

for each(var a in x..a) {
    var b = a.*::b
}
```

sbraconnier:
This option is set programmatically [here](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/util/JavaScriptSharedUtil.java#L206) in the OIE source code. I would also set the `indent_with_tabs` to `true` as it seems to be the default behavior of the editor.

At the moment, the function [JavaScriptSharedUtil#prettyPrint](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/util/JavaScriptSharedUtil.java#L143) is only used to pretty print the read-only "Generated Script" tab of a transformer step (or filter rule). It could be easily used as an action to manually format the code. I've already started working on a PR.

PR #90, opened 2025-05-17
Add the ability to manually format Javascript code (#87)

sbraconnier:
Default format shortcut is Ctrl+Shift+F (like Eclipse). I hesitated between Eclipse (Ctrl+Shift+F), Intellij (Ctrl+Alt+L) and Postman (Ctrl+B). 

Note that I also add the menu item "toggle comment" option to the new contextual "code" menu.


tonygermano (dismissed):
These changes all look good to me. I split the original commit into two because while adding the ability to toggle comments via the context menu makes sense, it is not related to the other change of being able to format the code. I also expanded the commit messages to better describe the changes.

As an example of what the code formatting can do, here is a before:
image

And after:
image

The context menu changes after both commits appears as the following:
image

gibson9583 (dismissed):

gibson9583:
@pacmano1 - noted same issue on malformed JS.  Only option around this would be to call validation first, but I don't think it's a deal breaker after kicking around with SC. Some beautifiers will and some won't format "bad" code.

pacmano1:
I came across this same issue with the content formatter. A different PR.  For example, both built in prettyPrints will format the "wrong" data type.  

So if I'm understanding you correctly, your position is the developer should know better.

pacmano1 (dismissed):

client/src/com/mirth/connect/client/ui/components/rsta/actions/FormatCodeAction.java
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (c) Mirth Corporation. All rights reserved.
+ * 
+ * http://www.mirthcorp.com
+ * 
+ * The software in this package is published under the terms of the MPL license a copy of which has
+ * been included with this distribution in the LICENSE.txt file.
+ */
```
kpalang: question: Is this Mirth Corp copyright intentional in a new file?

kpalang (changes requested):

client/src/com/mirth/connect/client/ui/components/rsta/actions/FormatCodeAction.java
sbraconnier: Well, not sure what the convention is for OpenIntegrationEngine new files. The PR was created before the branding changes, which removed (I think) the `license-header.txt` that was living at the root of the project. Every java source file had that header at that time.

kpalang: Gonna summon @tonygermano as the guy who knows about this stuff.

tonygermano: Good catch. I replaced it with a SPDX Header and named Simon as the copyright holder. NextGen should not be given copyright on any original files created by this project.

tonygermano (approved):

mgaffigan (approved):
Notes:
- Formatting fails pretty hard with `var x = <{a} {a}={a} />;`
- The user caret is moved to the start of the formatted selection (rather than wherever it was prior to formatting)
- The editor in general has some odd behavior with characters outside of the BMP

I don't think any of those are cause for correction.  LGTM

kpalang (approved):

sbraconnier-op3c:
[celebrate]     Simon Braconnier reacted to your message:
________________________________

PR #91, opened 2025-05-17
use semver library to avoid array comparisons

rogin:
Using library https://github.com/zafarkhaja/jsemver

Java 8 compatible


rogin:
Making a mess trying to merge my changes. Closing.

PR #93, opened 2025-05-17
Fix intermittent ConcurrentModificationException (#92)

### LINKED ISSUE #92: [BUG] Intermittent ConcurrentModificationException using getMergedConnectorMessage
**Describe the bug**
I personally experienced intermittent ConcurrentModificationException on heavy loaded channel. There is already some issues about this on the NextGen repo (see [ConcurrentModificationException in global post processor](https://github.com/nextgenhealthcare/connect/issues/6409) and [ConcurrentModificationException when using JS Writer](https://github.com/nextgenhealthcare/connect/issues/6123)).

**To Reproduce**
It's hard to reproduce, but you can visit the provided mirth issues for more details.

**Expected behavior**
`Message.getMergedConnectorMessage` should be thread safe.

**Actual behavior**
From time to time, a `ConcurrentModificationException` occurs.

**Screenshots**
NA

**Environment (please complete the following information):**
NA

**Workaround(s)**
There is no workaround.

**Additional context**
I'm is-simon now with my personnal github user (not working at "is" anymore).

I would be happy to work on this one. Should be easy enough...

sbraconnier:
Fix Intermittent ConcurrentModificationException using getMergedConnectorMessage (#92)


jonbartels:
@sbraconnier - I see what the change does and I see how it could prevent a concurrent modification exception. 

Can you link to an issue or discussion or other info about why or when the issue happens?

tonygermano (changes requested):
I think this may not actually fix the issue. I can't find any indication that this method would be called concurrently from two threads. The stack traces from the two nextgen issues don't have mirth versions on them, but they indicate that the problem occurs on one of these two lines (the line numbers shifted in mirth 4.4.0):
https://github.com/OpenIntegrationEngine/engine/blob/75ca77c111abf89ee8c6303abc3dc1ed6d0b767d/donkey/src/main/java/com/mirth/connect/donkey/model/message/Message.java#L155-L156

The map being modified concurrently is the argument to the `putAll` method. Since this method doesn't update that map, that means there is probably another thread that isn't trying to call this same method, but is instead updating the `responseMap` or `channelMap` for the same connectorMessage while this method is trying to read it.

sbraconnier:
Hmm you're right. I think it would be nice to let the synchronized block since the method is building the `mergedConnectorMessage` field and it will ensured that 2 threads won't initialized it concurrently. I'll take a look at the `putAll` calls.

tonygermano:
My guess without digging into it too much is that a destination queue thread is updating either the responseMap or the connectorMap while the post-processor is trying to build the mergedConnectorMessage.

It may be that just setting a couple retry attempts within this method if a ConcurrentModificationException is thrown would take care of it since this is pretty rare, and that wouldn't introduce the performance penalties associated with synchronization between threads.

mgaffigan (changes requested):
I know the ostrich algorithm is a thing, but this seems like a bad idea to try again.  `ConcurrentModificationException` is a signal we're violating a collection's synchronization contract.  It's being "nice" and telling us that so we can find where we went wrong.

If we don't want to synchronize, I'm sure there's a lock-free collection we could use, but I don't imagine that's required.

Why are there multiple threads involved in message processing?  I thought each channel thread serially processed messages.

tonygermano:
This is obviously a case of something being accessed concurrently that isn't expected to be accessed concurrently. For reference, the description for the  [HashMap](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/HashMap.html) class provides the following information, which I expect applies to how the `putAll` iterates over the argument's EntrySet.

> The iterators returned by all of this class's "collection view methods" are _fail-fast_: if the map is structurally modified at any time after the iterator is created, in any way except through the iterator's own remove method, the iterator will throw a [ConcurrentModificationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ConcurrentModificationException.html). Thus, in the face of concurrent modification, the iterator fails quickly and cleanly, rather than risking arbitrary, non-deterministic behavior at an undetermined time in the future.

tonygermano:
Closing per discussion on this PR and issue #92 .

PR #95, opened 2025-05-20
Install4j configuration

kayyagari:
I have commented out the line that included modules to make the binary start with Java8 by default(because our first release still targets Java8) , let me know if there are any objection to that.


server/docs/LICENSE.txt
jonbartels, line 1: Can this be a symlink or other pointer to the license at https://github.com/OpenIntegrationEngine/engine/blob/main/LICENSE ?

jonbartels (changes requested):

server/basedir-includes/oieserver.vmoptions
```diff
@@ -4,7 +4,7 @@
 -include-options conf/base_includes.vmoptions
 
 # Comment out the following line if you are running on Java 8
--include-options conf/default_modules.vmoptions
+#-include-options conf/default_modules.vmoptions
```
tonygermano: Java 8 should not be the default option.

server/basedir-includes/oieservice.vmoptions
```diff
@@ -4,7 +4,7 @@
 -include-options conf/base_includes.vmoptions
 
 # Comment out the following line if you are running on Java 8
--include-options conf/default_modules.vmoptions
+#-include-options conf/default_modules.vmoptions
```
tonygermano: Java 8 should not be the default option.

server/docs/LICENSE.txt
tonygermano: The main thing is that the build system currently needs the file at this location, and I'd be concerned about using symlinks when building on windows, but maybe it would be fine.

LICENSE
```diff
@@ -1,6 +1,6 @@
-Copyright (c) NextGen Healthcare. All rights reserved.
+Copyright (c) 2025 Open Integration Engine Contributors. Visit https://openintegrationengine.org/
 
-https://www.nextgen.com/products-and-services/integration-engine
+Copyright (c) 2005-2024 NextGen Healthcare.
```
tonygermano: I think the copyright information can be removed from the license completely so that it is just a clean copy of the MPL 2.0 license. The main place that copyright info should be specified (and it already is, in most cases) is the header on each file.

There is other MPL licensed software (like Rhino) which ships with the software and is not copyright NextGen or OIE contributors.

tools/install4j/oie-installer-config.install4j
```diff
@@ -0,0 +1,750 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<install4j version="11.0.3" transformSequenceNumber="11">
+  <directoryPresets config="../governance/branding/icons" />
+  <application name="Open Integration Engine" applicationId="4145-9206-7630-8076" mediaDir="${compiler:installer:mediaRoot}/server/build" shortName="oie" publisher="Open Integration Engine Project" publisherWeb="https://openintegrationengine.com" version="4.5.3" backupOnSave="true" autoSave="true" macVolumeId="88477e584eb462ba" javaMinVersion="1.8.0">
+    <variables>
+      <variable name="installer:mediaRoot" value="/Users/dbugger/projects/engine" />
+    </variables>
+    <jreBundles jdkProviderId="Zulu" release="8/8u452">
```
tonygermano: I was against bundling a JVM as it makes the download much larger, and people may not even want to use the bundled version, but if we're going to include one, I think it should be java 17.

tonygermano (changes requested):

tools/install4j/oie-installer-config.install4j
kayyagari: This config is for the installation binary to run, not for bundling with the binary.

server/basedir-includes/oieservice.vmoptions
kayyagari: My thinking here is that, we say v4.5.3 is compatible with Java8 but it doesn't run with Java8 unless user modifies this file after installation.

If we decide Java8 should not be the default option then I can make the minimum required version to 9 in install4j configuration.

LICENSE
kayyagari: Done.

server/docs/LICENSE.txt
kayyagari: Instead of symlinks I have updated both `LICENSE` and `server/docs/LICENSE.txt` files.

server/basedir-includes/oieservice.vmoptions
tonygermano: I'm not sure how this translates to install4j options, but java 8 is the minimum version to run the server. I think there was consensus that java 17 should be the recommended and default. I would say the latest LTS release, but I don't know how well we've tested it on java 21. Java 17 has been what is used in the NG docker containers for the past couple years, so we know that's well tested.

kayyagari: Right now Install4j is configured to *not* create new vmoptions files and the existing vmoptions files are the ones used by `oieserver` and `oieservice`.

These lines will be uncommented once we update to a version >= java 9, which I think will be likely in OIE v4.5.4 (or any other that comes after current 4.5.3).

server/docs/LICENSE.txt
tonygermano: I separated these changes out to a separate PR #99 so they do not need to be part of this one.

server/basedir-includes/oieservice.vmoptions
tonygermano: The lines should be uncommented in the repo and packages always, by design. It should work without changes for those running java 9+. The users running java 8 will need to comment the line out themselves post-installation.

tonygermano (approved):

jonbartels (approved):

PR #100, opened 2025-05-26
Migration to Gradle (new)

### LINKED ISSUE #52: [IDEA] Introduce Dependency Management (e.g. Gradle or Maven)
(fill in later)

NicoPiel:
Second attempt at a migration from Ant to Gradle. Compiles and start successfully.

Use `./gradlew dist -x test` (Donkey tests still fail)

Fixes #52


utilities/build.gradle.kts
jbeckers, line 1: all of utilities/* seems to be added by mistake?

jbeckers (commented):
Looks good to me

I'm curious what the diff would be between the official mirth 4.5.2 distribution and the result of this gradle build. I'll see if I can find some time to experiment this weekend.

.sdkmanrc
jonbartels, line 1: Should sdkman files be committed?

.vscode/settings.json
jonbartels, line 1: VSCode or IntelliJ or Eclipse project files should not be committed

utilities/src/main/java/org/example/utilities/JoinUtils.java
jonbartels, line 1: What are these generated files? Do they have to be in our source repo?

jonbartels (changes requested):
I left some trivial comments on project files and on the generated Java source files from Gradle.

I will need to do some builds and think about how to test that Gradle is producing a correct OIE build.

.sdkmanrc
tonygermano: Yes on this one. Kaur added it, and it helps ensure a more consistent developer environment.

.vscode/settings.json
tonygermano: agree

NicoPiel: agree

utilities/src/main/java/org/example/utilities/JoinUtils.java
NicoPiel: I don't think so, will get rid of them.

pacmano1:
@NicoPiel close this?

Issue #40, opened 2025-04-16
Create OIE Docker Image


jonbartels:
Create a Docker image for OIE.

The Mirth image is here - https://github.com/nextgenhealthcare/connect-docker

The env var support from this image is excellent and should be kept. 

This Image does some wierd things with SSL and downloading the Mirth artifacts. Evaluate if that needs to be kept or changed.

BridgeLink has some conversation about ideas and features at https://github.com/Innovar-Healthcare/BridgeLink/issues/4

Acceptance criteria:
* "mirth" naming conventions should be removed 
* breaking changes between connect-docker and OIE should be documented 
* provide a preferred JVM/JRE but allow for other implementations
* the `_MP` env var pattern support should be kept, changing `MP` to `OIE`
* the `LICENSE_KEY` env var should be removed
* publishing to https://hub.docker.com/u/openintegrationengine would be ideal, but is not required and can be ticketed separately

tonygermano:
> * the `_MP` env var pattern support should be kept, changing `MP` to `OIE`

I'm not sure if this is appropriate. `MP` stands for a generic value in **m**irth.**p**roperties. Fixing https://github.com/nextgenhealthcare/connect-docker/issues/12 would require specifying both the generic property and the correct file to which it should be applied.

Providing support for files in appdata via environment variables (server.id, keystore.jks, extension.properties) would also be good additions.

mgaffigan:
> My preference for this would be to mimic the nextgen setup and have the docker build happen in a fork of the NG docker repo, but with improvements (such as using a multi-stage build and occasionally refreshing our images with a new base image in addition to the changes mentioned above.)

Can you help me understand what the intent is?  This greatly complicates the docker story since it adds the distribution URL dependency, and makes dev-build separate from release-build.  Dockerfiles and OCI release artifacts are often used as complete alternatives to other package management and release channels.

> I think the official docker container should be reproducable using official releases of the application rather than building from source (the artifacts should already be signed with a real signature.)

Totally agree that reproducibility is important.  Dockerfile and containerization when with appropriate specification of dependency versions are great for making binary reproducibility from source to production artifact.  Assuming signing is still in the OIE release story, adding support should be relatively trivial since keys can be handled as [build secrets](https://docs.docker.com/build/building/secrets/).  Is there existing production signing machinery in the ant build?  The resultant OCI release artifact could contain the signed artifacts, but have no access to the original keys.

[...]

PR #101, opened 2025-05-27
Added docker build and quick-start

mgaffigan:
While this might be somewhat responsive to #40, it is primarily just to provide a quick-start way for a developer to build and test.  While trying to address Issue 5 I had hours of challenges getting things building happily - and shifted to docker as an escape hatch.

I do not expect this PR is ready to merge as is, but wanted to get something posted publicly.  Known issues include:

- [x] Not responsive to the acceptance criteria listed in #40
- [x] Uses a [personal container registry namespace](https://github.com/users/mgaffigan/packages/container/package/oie), rather than https://hub.docker.com/u/openintegrationengine
- [x] I'm not confident that I've implemented vmoptions handling correctly
- [x] Missing examples for docker-compose and playwithdocker

That said, it does work to the point of passing tests, running server, client, and simple message processing.

I'm new to the project, so if I've missed a step please let me know.


mgaffigan:
@tonygermano, I'm no longer aware of issues outside of the registry name (presumably someone owns the official URL?).

Differences I'm aware of relative to the original repo:

1. Uses multi-stage build to allow forks and keep dev, test, and prod on the same build pipeline and artifacts
2. Still unclear on jar signing story
3. Omitted [tests](https://github.com/nextgenhealthcare/connect-docker/tree/master/test) since they do not seem to run on any modern version of python and need to be reworked
4. Publish file is only pushing Temurin JDK on Ubuntu as ${tag}-jdk, and Temurin JRE on Alpine as ${tag} (e.g.: latest-jdk, latest-jre)

Happy to discuss as needed.  Every compose file, kubernetes deployment, etc... that I've tested doesn't notice the difference between this and stock.

tonygermano:
@mgaffigan  I pulled your launcher script out of this PR, made significant changes, and put it in its own PR in #119 . I think it is useful outside of a docker context, and it can still be used with docker. Would you mind reviewing it and giving any feedback?

mgaffigan:
@tonygermano, Excellent!  I'd love to see that merged and agree it is useful outside of docker.

.dockerignore
kpalang, line 1: suggestion: I think having an "ignore everything" and "unignore this" approach would be more readable here
```
*
!setup/
```

README.md
```diff
@@ -19,6 +20,12 @@
 - [Licensing](#licensing)
 - [Acknowledgments](#acknowledgments)
 
+---
+## Quick-start
+
+Start an instance of OIE with `docker run -p 8443:8443 ghcr.io/mgaffigan/oie` and
```
kpalang: issue: We have an `openintegrationengine/engine` docker hub repo available. I'll help setting up the build pipeline once the Dockerfiles are merged.

But also, I wouldn't include this line in the readme before be have an actual image pushed.

connectservice/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:18
```
kpalang: suggestion: I'd use a newer node, 22 is latest LTS.

connectservice/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /app
+
+COPY package*.json ./
+RUN npm install
```
kpalang: thought: If you have time and interest, look into the new hotness that is [bun](https://bun.sh/) ;)

docker/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+name: OIE-release
+
+services:
+  jre:
+    image: ${repo}:${tag:-dev}
+    build:
+      context: ../
+      dockerfile: Dockerfile
+      target: jre-run
+      x-bake:
```
kpalang: suggestion: `platforms` seems to be a [standard feature](https://docs.docker.com/reference/compose-file/build/#platforms) in compose spec. The `x-bake` might not be needed.

.gitattributes
```diff
@@ -0,0 +1 @@
+*.sh text eol=lf
```
jonbartels: This is OK since there are not any .sh files in the repo yet

.gitignore
```diff
@@ -11,6 +11,12 @@
 hs_err_pid*
 replay_pid*
 
+##############################
+## Javascript
+##############################
+node_modules/
```
jonbartels: This is OK to have in gitignore, but are we really introducing node.js for an image build?

DEVELOPERS.md
```diff
@@ -0,0 +1,33 @@
...
+## Build Environment
+
+To build the solution, you must have a Java 1.8 JDK+FX and Apache Ant.  This
+can be installed by [sdkman](https://sdkman.io/) by executing `sdkman env install`.
+
+## Build Process
+
+From the `server/` directory, run `ant -f mirth-build.xml -DdisableSigning=true`.
+
+If you are using Mirth Connect Administrator Launcher, you may need to omit
```
jonbartels: MCAL may also need the `-k -d` flags on start to ignore self signed certs

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+WORKDIR /app
+# sdkman requires bash
+SHELL ["/bin/bash", "-c"]
+
+# Stage 1a: Install dependencies
+# Install necessary tools
+COPY .sdkmanrc .
+RUN apt-get update\
+    && apt-get install -y zip curl\
+    && curl -s "https://get.sdkman.io?ci=true" | bash \
```
jonbartels, line 23: Why do we need SDKMan and not just an ubuntu jdk image?

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+
+# Stage 1b: Build the application
+# Copy the entire source tree (excluding .dockerignore files), and build
+COPY --exclude=docker . .
+WORKDIR /app/server
+RUN source "$HOME/.sdkman/bin/sdkman-init.sh" \
+    && ANT_OPTS="-Dfile.encoding=UTF8" ant -f mirth-build.xml -DdisableSigning=true
+
+# Stage 2b: JDK runtime container
+FROM eclipse-temurin:21.0.7_6-jdk-noble as jdk-run
```
jonbartels: Can this just specify `21` without the minor and patch version details?

docker/mirth-connect.sh
jonbartels, line 1: Is this script suitable for all builds or only ones initiated from docker?

connectservice/.dockerignore
jonbartels, line 1: Why a second .dockerignore? Does the parent directory ignore file apply here?

connectservice/Readme.md
jonbartels, line 1: Move connect service to a separate PR

connectservice/Readme.md
```diff
@@ -0,0 +1,21 @@
+# Connect Service
+
+Connect Service is a sample implementation of a web service which records usage
+of Mirth Server, and provides access to a notifications feed. The service implements
```
jonbartels: OIE not Mirth :D

jonbartels (changes requested):

mgaffigan:
@jonbartels,

Thanks for the review.  I think this PR got out of date and looks like it ended up with an accidental merge from #114 at some point.  Let me clean it up to minimums.  I'll pull the docker-compose examples out as well, assuming they fit better with #126.

connectservice/Readme.md
mgaffigan: Accidental merge - does not belong in this PR.  I'll take this comment as feedback for #114.

connectservice/Readme.md
mgaffigan: Agree. Accidental merge - does not belong in this PR.

connectservice/.dockerignore
mgaffigan: Accidental merge - does not belong in this PR.  I'll take this comment as feedback for #114.

docker/mirth-connect.sh
mgaffigan: This script was the basis for #119.  It should not be duplicated for docker.  I'll update to assume #119 gets merged before this PR.

Dockerfile
mgaffigan: It can, but I disagree with it philosophically since then a particular git rev is not guaranteed to produce the same behavior.  Goal is to eliminate "it works on my machine".  I'd suggest something like [Dependbot](https://dev.to/oracle2025/how-to-keep-a-dockerfile-updated-with-dependabot-1mdn) be used to submit a PR when new versions are available.

If the project model is to "use latest", let me know and I will update it.

Dockerfile
mgaffigan: I think either is acceptable, and avoiding sdkman would certainly make the dockerfile simpler.  I think I went this route because the zulu jdk image does not contain ant, so I had to pull additional packages regardless.

I'll see what that would look like.

.gitignore
mgaffigan: Accidental merge - does not belong in this PR. I'll take this comment as feedback for https://github.com/OpenIntegrationEngine/engine/pull/114.

.gitattributes
mgaffigan: I'll remove this since I separately opened #111 to address this issue.

DEVELOPERS.md
mgaffigan: Updated.  Feel free to send suggested language if I didn't get it right.

.dockerignore
kpalang, line 1: suggestion: I think the "ignore everything but not this" approach would be more readable here.

For example:
```
*
!setup/
```

Dockerfile
kpalang, line 1: I would split the Dockerfiles into two separate ones depending on where they source the `engine/` directory.

Firstly, I quite dislike build containers as I've had too many issues with layer caching platform-specific optimizations. With a long-building application like OIE it starts to matter. I'm not totally against it, but in my opinion it shouldn't be the recommended way to build.

The first one, let's say `local.Dockerfile` would copy the `setup/` directory that has been build with `ant`. And the second one, `prod.Dockerfile`, would download a release from GitHub. In my opinion both of the resulting images should be based on existing Java images (Zulu, Temurin, etc.).

Both of them should be templated to allow for different base image tags (jre vs jdk).

Dockerfile
```diff
@@ -0,0 +1,98 @@
+# syntax=docker/dockerfile:1.7-labs
```
kpalang: suggestion: I think we should stay on `docker/dockerfile:1`

DEVELOPERS.md
```diff
@@ -0,0 +1,34 @@
...
+
+## "Easy Path" with docker
+
+```bash
+# Build using docker
+docker build -t oie-dev .
+# Start an ephemeral image
+# NOTE: All data will be deleted on stop due to --rm.  Use a volume for "real" use.
+docker run --rm -p 8443:8443 oie-dev
+```
```
kpalang: issue: I am against recommending building in Docker as the first approach. Layer caching will become an issue quite soon.

There are two ways to break layer caching:
1. Editing `Dockerfile` to invalidate cache and pull in the newly build directory
    - Way too tedious to be considered acceptable
2. Setting `--no-cache` when building the image
    - Forces rebuilding of _every_ layer and costing unnecessary time

I think the recommended approach should be to use SDKMAN to get the dependencies and leave build containers to people who know what they're doing.

DEVELOPERS.md
```diff
@@ -0,0 +1,34 @@
...
+
+Then use [Ballista](https://github.com/kayyagari/ballista) to connect to 
+https://localhost:8443/ and login using admin admin.
+
+## Build Environment
+
+To build the solution, you must have a Java 1.8 JDK+FX and Apache Ant.  This
+can be installed by [sdkman](https://sdkman.io/) by executing `sdkman env install`.
```
kpalang: question: Isn't the command just `sdk` instead of `sdkman`?

DEVELOPERS.md
```diff
@@ -0,0 +1,34 @@
...
+If you are using Mirth Connect Administrator Launcher, you may need to omit
+`-DdisableSigning=true` to support JWS signatures and run MCAL passing `-k -d`
+to make it ignore self-signed certificates. Launchers like 
+[Ballista](https://github.com/kayyagari/ballista) do not require signing, and 
+signing adds considerable time to the build process.
+
+## Run
+
+After build, run the server by invoking `server/mirth-server-launcher.jar`.  An
+example of how to do this is listed in `docker/mirth-connect.sh`.
```
kpalang: issue: Now non-existant reference.

I assume the reference was correct in the previous, larger version of the PR.

kpalang (changes requested):

README.md
mgaffigan: Yep.  I pulled this line out of this PR assuming it will land with 126.

connectservice/Dockerfile
mgaffigan: Accidental merge - does not belong in this PR. I'll take this comment as feedback for https://github.com/OpenIntegrationEngine/engine/pull/114.

connectservice/Dockerfile
mgaffigan: Accidental merge - does not belong in this PR. I'll take this comment as feedback for https://github.com/OpenIntegrationEngine/engine/pull/114.

.dockerignore
mgaffigan: Since we are building from source, I think an exclude list would be substantially more manageable than an include list.

docker/docker-compose.yml
mgaffigan: This is removed in the latest rev of the PR.  I'm assuming this will land in 126 for a release that pull signed binaries.

.dockerignore
mgaffigan: Closing as duplicate.

Dockerfile
mgaffigan: Assuming we can't release from an OCI image, then I agree with having a separate dockerfile for running golden bits.  This dockerfile is specifically targeted at developer run, or prod run assuming that we're building inside docker.

Dockerfile
mgaffigan: `docker/dockerfile:1` doesn't support `COPY --exclude=...` and would thus result in increased image sizes.

DEVELOPERS.md
mgaffigan: I'm not sure I follow your comment about layer caching.  That's a feature - not a bug.  `sdk env install` will be re-run any time `.sdkmanrc` is changed.  What cached resource would we not want to keep on build?

DEVELOPERS.md
mgaffigan: It is.  Updated.

DEVELOPERS.md
mgaffigan: Yep.  Updated.

Dockerfile
mgaffigan: Also, re: long-building, I'm at 22 seconds as unless I run tests and signing.  I think that would go down pretty significantly with incremental builds.

Dockerfile
jonbartels: Your code is correct. I forgot about guaranteed build contracts.

Dockerfile
mgaffigan: See https://github.com/mgaffigan/oie/commit/123bbe24f14cb132dd1474d11635fb0213b1da43 for changes to avoid sdkman.  Problems are:

- Zulu nor Temurin seem to publish a JDK with FX included.  Install from deb or tgz requires quite a few deps.
- Ant requires separate installation
- Docker revisions can get out of sync with `.sdkmanrc`

I don't love sdkman, but it does seem simpler to use it in the builder container.

.dockerignore
kpalang: When writing the previous comment, I was thinking of the release container.
But, if we're talking about a build container for development, I still think includelisting the 6 directories + possibly `.sdkmanrc` might be the cleaner way

Dockerfile
kpalang: I agree with the opinion that neither of the two options are perfect.  Using a jdk image can easily drift with our preferred version from `.sdkmanrc`, and installing sdkman in a container is just nasty.

But in this case installing sdkman in the container is the better way to go to ensure as identical as possible envs.

>Zulu nor Temurin seem to publish a JDK with FX included. Install from deb or tgz requires quite a few deps.

I'm not sure this is true. the `8.0.442.fx-zulu` specified in `.sdkmanrc` is one example of that. Unless you're talking about Docker images.

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+    /app/server/setup ./
+
+VOLUME /opt/engine/appdata
+VOLUME /opt/engine/custom-extensions
+
+EXPOSE 8443
+
+USER engine
+ENTRYPOINT ["./configure-from-env.sh"]
+CMD ["./oieserver"]
```
kpalang, line 99: question: These images have been moved to `engine-docker` and will be built with https://github.com/OpenIntegrationEngine/engine-docker/pull/1. Would it be reasonable to edit the `Dockerfile` in this repo to be a build/local container only?

server/basedir-includes/configure-from-env
kpalang, line 1: issue: @OpenIntegrationEngine/maintainers correct me if I'm wrong, but this file might live in [engine-docker](https://github.com/OpenIntegrationEngine/engine-docker) repo instead. I don't have a good suggestion on how to sync the two.

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+WORKDIR /app
+# sdkman requires bash
+SHELL ["/bin/bash", "-c"]
+
+# Stage 1a: Install dependencies
+# Install necessary tools
+COPY .sdkmanrc .
+RUN apt-get update\
+    && apt-get install -y zip curl\
+    && curl -s "https://get.sdkman.io?ci=true" | bash \
```
kpalang, line 23: nitpick: Downloading and executing arbitrary scripts is not good practice. The correct way should be to download the binaries from GH and set them up ourselves, but it quite big hassle and I've never completed it, so might be good enough for now.

Dockerfile
kpalang: question: Could the `--exclude` directives be run as `rm`-s in the `builder` stage?

DEVELOPERS.md
kpalang: What I mean was that layer-caching is good on some scenarios, but not necessarily others. What I fear is going to happen is changes in engine code do not trigger a cache miss on the `COPY --exclude=docker . .` layer.

I am testing this right now.

kpalang: Looks like changes in code do indeed trigger a cache miss.

Dockerfile
mgaffigan: You cannot run an `rm` in the same layer as a COPY, so the image size would still be inflated.  I think the alternative here would be to add an option to `mirth-server.xml` to not copy them to `setup`.

The other options I'm disagreeing with are:
1. Copy each item from setup by name (verbose, fragile if new item added)
1. Copy to a third location (symlink?), then copy to final (ew for complexity)
1. Take the hit on layer-size (48 MB)
1. ADD with special tar (ew for time/space and general disgust)

I generally agree with the advice to stay on `docker/dockerfile:1`, but `--exclude` is fairly well supported by this point, and *should* get [taken out of labs soon](https://github.com/moby/buildkit/issues/2853#issuecomment-2855262799).

Dockerfile
mgaffigan: Resolving since https://github.com/OpenIntegrationEngine/engine-docker/pull/1 merged.  Please reopen if there is something unaddressed.

Dockerfile
kpalang: Out of curiosity since I don't currently have the means to test this, but when doing a multi-stage build with a `COPY --from`, will the layers from the first stage be included in the second stage or is there any squashing action going on?

.dockerignore
mgaffigan: Can you help me understand what you are intending?  The primary goal of a `.dockerignore` is to avoid copying build artifacts, not to select things for the docker build.  It should generally match .gitignore (since if you are not putting it in code-control, it probably should not be used for build).

Dockerfile
mgaffigan: I'm referring to docker images.  Azul and Temurin publish tgz of JavaFX, but they do not provide pre-built images.  See https://github.com/mgaffigan/oie/commit/123bbe24f14cb132dd1474d11635fb0213b1da43 for the changes to install from tgz rather than sdkman.

Dockerfile
mgaffigan: I'm not sure I see the difference between using sdkman and installing sdkman - since they both download and run untrusted scripts.

If we were aiming for zero-trust, then we would need to shift to installing from tgz and add hash verifications.  sdkman is fundamentally opposed to that.  Practically, I would recommend separately maintaining and publishing a javafx container then basing this file on that.

Dockerfile
mgaffigan: Can you clarify what you are suggesting be removed?  I think it is useful to be able to do a local build of ubuntu or alpine, since those are the release environments and would thus both be needed for testing.

.dockerignore
kpalang: The way I see it is that we have two options to use containers in the development cycle.
1. Compile OIE natively (`ant build`) from the host machine and package the build result into a container for runtime testing
2. Compile OIE inside a container that already packages required dependencies _(and continue building runtime images from this)_

With both approaches the build context should be as small as possible.
With approach 1 the context should only contain the `setup/` directory and possibly some individual files. This is what I intended in the first comment.
With approach 2 things get a little more complicated, as we might need most of the dirs + files in the repo par some license or instructions files. But it is quite easy to find us with a bunch of little files that tweak some tooling or another, but are not necessary to compile OIE, which is why in my mind whitelisting only the files + dirs required for compilation might be the simpler way to go.

Dockerfile
kpalang: In my mind the difference is that we have no control over the script being downloaded from `https://get.sdkman.io`. It carries the risks of DNS poisoning (although highly unlikely), and malicious changes by sdkman maintainers themselves - also highly unlikely given their size. But then again, `xz` was even larger.

What I'm saying is this is something we should keep in mind. I don't have a viable alternative right now.

server/basedir-includes/configure-from-env
mgaffigan: I copied this file from engine-docker/entrypoint.sh. If it is present here, it doesn't need to be present in the release image repo.

It seems generally useful, even if not used with docker.

.dockerignore
mgaffigan: This PR is exclusively addressing the containerized build of OIE.  I think the engine-docker repo already addresses the creation of a container from artifacts produced outside of docker.

This should be the logical dual of .gitignore.  If it is in git, it needs to be in the docker context.

Dockerfile
mgaffigan: @kpalang, is this a blocker for you?  I don't see an easier way here.  See the other branch for the alternative.

Dockerfile
mgaffigan: @kpalang, is this a blocker for you?

Dockerfile
kpalang: Not really. I guess this is fine.

Dockerfile
kpalang: No.

The original question was most definitely an important one, but I seem to not remember what the reason for it was.

Dockerfile
```diff
@@ -0,0 +1,96 @@
+# syntax=docker/dockerfile:1
```
kpalang: praise: Love the syntax inclusion here

kpalang (dismissed):

DEVELOPERS.md
```diff
@@ -0,0 +1,30 @@
+# For developers and contributors
+
+## Build and run locally
```
jonbartels: Java 8, a JDK with some extra gubbins, Ant.. is this a good case for a build container? 

I'm not too familiar with them but my understanding is that a build container can be used to build and compile a project without farting around with a lot of dev tool or dependency installs. Its also cross platform.

This idea is likely a separate change from this PR.

DEVELOPERS.md
```diff
@@ -0,0 +1,30 @@
...
+
+To build the solution, you must have a Java 1.8 JDK+FX and Apache Ant.  This
+can be installed by [sdkman](https://sdkman.io/) by executing `sdk env install`.
```
jonbartels: IIRC SDKMan isn't supported on windows. Do you need separate *nix, mac, and winderz instructions?

Dockerfile
jonbartels: What does syntax do?

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+
+# Stages:
+# 1. Builder Stage: Compiles the application and resolves dependencies.  Produces
+#    JAR files that can be deployed.
+#      1a. Install dependencies
+#      1b. Build the application
+# 2. Runner Stage: Creates a lightweight image that runs the application using the JRE.
+
+FROM ubuntu:noble-20251013 AS builder
```
jonbartels, line 12: Why not `latest` - this is now 2 months out of date

Dockerfile
mgaffigan: Stages are separate unless a subsequent stage references a prior stage in a "FROM". 

None of the builder stage ends up in the deployment artifact, except that which is specifically COPY'd over.

mgaffigan: --exclude got released, so this is no longer an issue.

Dockerfile
mgaffigan: @jonbartels, It sets the minimum required version of docker to interpret the file.  The `--exclude` argument for the copy command was only added in a later syntax.

DEVELOPERS.md
mgaffigan: This dockerfile contains a build stage, unless you are referring to a dev container.  I do think adding a dev container would be nice, but I haven't done that.

This PR:
1. From any docker host, will build the repo using sdkman and ant
2. Will copy the build artifacts to a new ubuntu or alpine stage, that can be published as an OCI artifact or ran locally

DEVELOPERS.md
mgaffigan: I think *nix and mac can go as is, but windows should be clarified, yes.  This also needs to be updated for the new Java 17 minimum build.

Dockerfile
mgaffigan: `ubuntu:noble-20251013` is current ubuntu LTS release.  Specifying a specific tag is considered best practice for reproducible builds (though admittedly that would be a hash, not a tag, when taken to heart).  Dependabot can submit PR's automatically to track latest, which keeps approval in the hands of the project.

I am unopposed to `latest` if preferred.

DEVELOPERS.md
mgaffigan: See https://github.com/mgaffigan/oie/tree/feature/add-devcontainer for a devcontainer.  Works fine locally, but for some reason archive.apache.org is blocking the github codespaces IP Addresses.

DEVELOPERS.md
mgaffigan: Updated language to address windows.

kryskool: FYI the cli interface for sdkman start to be [rewrite in Rust ](https://github.com/sdkman/sdkman-cli-native) and also available on Windows.

Dockerfile
tonygermano: My preference is to keep the named tags and use dependabot to bump tags.

Dockerfile
```diff
@@ -0,0 +1,96 @@
+# syntax=docker/dockerfile:1
+
```
tonygermano, line 4: chore: include license header after syntax declaration

DEVELOPERS.md
tonygermano: Only the `-k` flag is necessary to allow self-signed certs.

The description for `-d` (long form `--allow-incorrect-digest`) is
> Allows JARs that fail digest verification.
This should only be used for
development/testing purposes.

I have never seen a case where `-d` was needed, and I am not sure of the use case for it. Maybe if the jar is replaced in the local cache but not on the server? I would not encourage usage of the `-d` flag without knowing what its function is.

server/basedir-includes/configure-from-env
tonygermano, line 1: suggestion (non-blocking): move script to a `bin/` directory instead of the base dir

While it can be called independently, I don't expect this to get much usage outside of docker since it is probably easier to edit the configuration files directly than to set up the environment variables in a typical interactive environment. You also would only want to use the script one time, since it modifies the config files by appending to them, and then use the regular launch script on subsequent runs so as to not keep modifying the files in an environment where the files are not reset to default prior to each launch.

I think putting it in a `bin/` directory makes it discoverable without cluttering the base dir. That directory can then be used for future tools or scripts as well and the chmod can likely be applied to the entire directory.

DEVELOPERS.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+## Connect
+
+Then use [Ballista](https://github.com/kayyagari/ballista) to connect to 
+https://localhost:8443/ and login using admin admin.
+
+If you are using Mirth Connect Administrator Launcher, you may need to omit
+`-DdisableSigning=true` to support JWS signatures and run MCAL passing `-k -d`
+to make it ignore self-signed certificates. Launchers like
```
tonygermano:
```suggestion
`-DdisableSigning=true` to sign the client jars and run MCAL from the cli
passing `-k` to make it ignore self-signed certificates. Launchers like 
```

Dockerfile
jonbartels: Dependa is good for me. I forgot it existed

Dockerfile
```diff
@@ -0,0 +1,96 @@
...
+COPY .sdkmanrc .
+RUN apt-get update\
+    && apt-get install -y zip curl\
+    && curl -s "https://get.sdkman.io?ci=true" | bash \
+    && source "$HOME/.sdkman/bin/sdkman-init.sh" && sdk env install \
+    && rm -rf /var/lib/apt/lists/*
+
+# Stage 1b: Build the application
+# Copy the entire source tree (excluding .dockerignore files), and build
+COPY --exclude=docker . .
```
tonygermano: question (non-blocking): Can this exclusion be moved to the .dockerignore file?

Does it actually exclude the .dockerignore file? It appears to be excluding a non-existent `docker/` directory (which I think I would prefer to have to hold all of the docker related files)

tonygermano (changes requested):
Neither of these comments are blocking:

1. This does not work in my environment (podman installed from package manager in debian bookworm) because of the `COPY --exclude` syntax. If I was using docker instead, I think it still would not work without some modification to my environment to tell it to use BuildKit.
2.  At some point the content of `DEVELOPERS.md` will need to be reconciled with the changes to `CONTRIBUTING.md` in #201 , and likely most of this "how to build" documentation should be moved to https://github.com/OpenIntegrationEngine/docs-website where it can be better organized rather than living in this repo.

server/basedir-includes/configure-from-env
mgaffigan: I agree with the idea, but we've got a lot of directories already.  At some point we should probably take a breaking change and reorganize.  Deferring till then.

Dockerfile
mgaffigan: Updated.

Dockerfile
mgaffigan: It was a no-op, removed.

DEVELOPERS.md
mgaffigan: Per discussion in Github, I updated to refer to the docs site, and only kept a reference to Ballista.

DEVELOPERS.md
mgaffigan: Per discussion in Github, I updated to refer to the docs site, and only kept a reference to Ballista.

mgaffigan:
@tonygermano 
> 1. This does not work in my environment (podman installed from package manager in debian bookworm) because of the `COPY --exclude` syntax. If I was using docker instead, I think it still would not work without some modification to my environment to tell it to use BuildKit.

I updated the syntax line to be more specific - hopefully that will give a better error in podman if --exclude is not supported.

> 2. At some point the content of `DEVELOPERS.md` will need to be reconciled with the changes to `CONTRIBUTING.md` in #201 , and likely most of this "how to build" documentation should be moved to https://github.com/OpenIntegrationEngine/docs-website where it can be better organized rather than living in this repo.

Agreed.  Updated DEVELOPERS.md to reference the docs website.  Not sure about reconciling 201. I wouldn't mind moving that reconciliation to #201, since I'm trying for a small PR here that delivers the dockerfile.

tonygermano (dismissed):

jonbartels (dismissed):

tonygermano (changes requested):
I preferred the short and sweet DEVELOPERS.md before Jon's changes. I don't think we're at the point yet to say the docker build is the preferred method. As I mentioned in my previous review, it doesn't work in my environment, and I plan to continue building without docker, but I wasn't going to block others from using it. I also expect the docker build to be slightly slower because it has to copy the source files into the container first and otherwise is running the same build. The local build will definitely be faster if we get incremental builds working eventually.

I think it's good to let the repo have just enough info and refer to the docs site, which can evolve more easily as things change.

mgaffigan:
I've removed developers.md entirely to leave that to a separate PR. While we opine about documentation, this PR nears its 1st birthday.  Perfect is the enemy of done.

This PR being merged is blocking for smoke testing in github pipelines, and frankly for many modern dev workflows.

tonygermano (approved):

NicoPiel (approved):

jonbartels (approved):
I have recently worked on Dockerfiles for my own deploys. This update covers updates to the entrypoint script that I have seen in the wild. The new dockerfile seems correct and applies some recent best practices

Issue #110, opened 2025-05-30
Add .gitattributes


mgaffigan:
If I'm missing something here, please say so, but it seems like there's not much consistency on whether the repo uses CRLF or whether it uses LF.  Most projects specify a .gitattributes to ensure the repo is stored in LF, and each contributor ends up with an appropriate EOL according to their system preference.

I don't think this is breaking anything, and it would be a huge commit, but it is likely worth doing at some point.

A minimal (still acceptable) .gitattributes would just be `* text=auto`, then run `git add --normalize .`

Here's the top file extensions and their formats as of today (output of `git ls-files --eol`):

| Extension   | i/crlf | i/lf | i/mixed | i/none | i/-text | Grand Total |
| ----------- | ------ | ---- | ------- | ------ | ------- | ----------- |
| .xml        | 6267   | 136  | 0       | 0      | 0       | 6403        |
| .java       | 569    | 1223 | 4       | 0      | 1       | 1797        |
| .jar        | 0      | 0    | 0       | 0      | 455     | 455         |
| .png        | 0      | 0    | 0       | 0      | 113     | 113         |
| .sql        | 44     | 53   | 0       | 11     | 0       | 108         |
| .form       | 0      | 62   | 0       | 0      | 0       | 62          |
| .js         | 3      | 38   | 0       | 2      | 0       | 43          |
| .txt        | 19     | 14   | 1       | 5      | 0       | 39          |
| .properties | 11     | 7    | 0       | 0      | 0       | 18          |
| .css        | 4      | 11   | 0       | 3      | 0       | 18          |
| .hl7        | 0      | 14   | 0       | 0      | 1       | 15          |
| .md         | 0      | 9    | 0       | 0      | 0       | 9           |
| .classpath  | 3      | 4    | 1       | 0      | 0       | 8           |
| .project    | 7      | 1    | 0       | 0      | 0       | 8           |
| .vmoptions  | 0      | 5    | 0       | 0      | 0       | 5           |
| .tld        | 0      | 5    | 0       | 0      | 0       | 5           |

[Full table](https://github.com/user-attachments/files/20527520/oie.eol.xlsx).

tonygermano:
[...]

PR #111, opened 2025-05-30
Added .gitattributes

mgaffigan:
Re: #110.  Should not be merged without `git add --renormalize .` commit.  For an example commit, see https://github.com/OpenIntegrationEngine/engine/compare/main...mgaffigan:feature/add-gitattributes-renormalize?expand=1&w=1

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,146 @@
+# Common settings that generally should always be used with your language specific settings
+
+# Auto detect text files and perform LF normalization
+*          text=auto
+
+#
+# The above will handle all files NOT found below
+#
+
+# Documents
+*.bibtex   text diff=bibtex
+*.doc      diff=astextplain
+*.DOC      diff=astextplain
+*.docx     diff=astextplain
+*.DOCX     diff=astextplain
+*.dot      diff=astextplain
+*.DOT      diff=astextplain
+*.pdf      diff=astextplain
+*.PDF      diff=astextplain
+*.rtf      diff=astextplain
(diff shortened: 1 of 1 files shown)


mgaffigan:
Updated to avoid requirement for large commit, also added six files which have mixed end-of-line within the same file.  Diff which ignores eol changes can be viewed at https://github.com/OpenIntegrationEngine/engine/pull/111/files?w=1

jonbartels (approved):

kryskool (approved):

tonygermano (changes requested):
The mixed eol commits that rewrite a significant part of the file are what I was originally concerned about in my comments in #110 . I've since learned about `.git-blame-ignore-revs`, which supposedly is supported by github, so I'm planning to test that out.

Assuming it does, then my concern about the git blame will not be an issue when we fully normalize. The only issue at that point is that we'd be creating merge conflicts for any sharing between forks if the other forks do not normalize as well.

gibson9583 (approved):

mgaffigan:
@tonygermano, I did look at git-blame-ignore-revs, but since it only applies in a non-default configuration, did not include it.

If the issue is limited to the 6-files, it is still valuable to merge without that to prevent further issues in the future.  The 6-file count is reduced from the original 600 that `git add --renormalize .` wanted.  If there's consensus on the 6-whitespace-only edits being too much, then I'm happy to revert them.

kpalang (approved):

PR #119, opened 2025-06-14
Add oieserver launcher script (bash version)

tonygermano:
This version of the script is a modified version of [Mitch Gaffigan's script](https://github.com/mgaffigan/oie/blob/a3709ea64ff9498dec74b7a815ca869646be57d5/docker/mirth-connect.sh), which in turn was a bash port of [my clojure app](https://github.com/OpenIntegrationEngine/engine/blob/0004d5afd2f4f4f1cb27db80f7395400a5287af2/tools/engine-launcher/src/org/openintegrationengine/engine/server/launcher.clj) from #61.

This script is the main launcher for the Open Integration Engine (OIE) server. It prepares the Java environment and executes the server launcher JAR file.

The script automatically finds a compatible Java runtime (version 17+ by default) by searching for a valid executable in the following priority order:
  1. The OIE_JAVA_PATH environment variable.
  2. The -java-cmd directive in the oieserver.vmoptions file or included .vmoptions files. Must specify the path to the 'java' executable. This is the preferred way to declare the desired version for running the server and can be overridden by OIE_JAVA_PATH. Can be a relative path from the location of this script.
  3. The JAVA_HOME environment variable.
  4. The 'java' command available in the system's PATH.

It also parses the 'oieserver.vmoptions' file to configure JVM options, system properties (-D...), and classpath modifications.

Issue: https://github.com/OpenIntegrationEngine/engine/issues/2


server/basedir-includes/oieserver
```diff
@@ -0,0 +1,229 @@
...
+
+JAVA_OPTS=("${VMOPTIONS[@]}" 
+           "-cp" "$CLASSPATH" 
+           "com.mirth.connect.server.launcher.MirthLauncher" 
+           "${APP_ARGS[@]}")
+
+# Launch Open Integration Engine (as this PID with exec)
+echo "Starting Open Integration Engine..."
+echo "Using Java from: $FINAL_JAVA_CMD"
+echo "$FINAL_JAVA_CMD ${JAVA_OPTS[*]}"
```
mgaffigan, line 280: This seems duplicative.  Perhaps the options are printed as `Options: ` or it's printed once as `Using Java command: $JAVA $OPTIONS`?

server/basedir-includes/oieserver
```diff
@@ -0,0 +1,229 @@
...
+        elif [[ "$line" =~ ^-java-cmd[[:space:]]+(.+) ]]; then
+            # Handle -java-cmd directive
+            local candidate_java_cmd="${BASH_REMATCH[1]}"
+            # Resolve path to an absolute one before validation
+            local resolved_candidate_cmd
+            resolved_candidate_cmd=$(resolve_absolute_path "$candidate_java_cmd")
+            
+            if [[ -z "$FINAL_JAVA_CMD" ]] && is_valid_java_version "$resolved_candidate_cmd" "the -java-cmd directive in '$file'"; then
+                VMOPTIONS_JAVA_CMD="$resolved_candidate_cmd"
+            fi
```
mgaffigan: I've spent more time troubleshooting things trying to be smart about recovering from bad circumstances than most any other cause.  If the user is specifying a java version, I'd suggest it either use that java version, or fail specifying it is incompatible.

I can understand using `is_valid_java_version` to check JAVA_HOME when no explicit JRE is specified, but this is presumably coming from custom.vmoptions.

For an "it just works" story, I suggest detecting and using a bundled JRE rather than trying to get clever.  Very few people will balk at an extra 150 MB, and those that will are probably building from source or using the tgz distribution.

server/basedir-includes/oieserver
```diff
@@ -0,0 +1,229 @@
...
+            CLASSPATH="${CLASSPATH}:${BASH_REMATCH[1]}"
+        elif [[ "$line" =~ ^-classpath/p[[:space:]]+(.+) ]]; then
+            # Handle -classpath/p directive (prepend to existing classpath)
+            CLASSPATH="${BASH_REMATCH[1]}:${CLASSPATH}"
+        elif [[ "$line" =~ ^-java-cmd[[:space:]]+(.+) ]]; then
+            # Handle -java-cmd directive
+            local candidate_java_cmd="${BASH_REMATCH[1]}"
+            # Resolve path to an absolute one before validation
+            local resolved_candidate_cmd
+            resolved_candidate_cmd=$(resolve_absolute_path "$candidate_java_cmd")
```
mgaffigan: I would presume this should be resolved relative to the file being parsed?

server/basedir-includes/oieserver
```diff
@@ -0,0 +1,229 @@
...
+# @echo: The resolved, canonical absolute path.
+resolve_absolute_path() {
+    local path_to_resolve="$1"
+
+    # Explicitly handle tilde expansion first
+    if [[ "$path_to_resolve" =~ ^~ ]]; then
+        # Use eval to expand the tilde to the user's home directory.
+        # This correctly handles "~", "~/path", and "~username/path".
+        path_to_resolve=$(eval echo "$path_to_resolve")
+    fi
```
mgaffigan: Is this beneficial?  This would mean that using the same script under two users would produce different results (e.g.: running as a user, then running from systemd).

server/basedir-includes/oieserver
```diff
@@ -0,0 +1,229 @@
...
+        fi
+    done < "$file"
+    return 0
+}
+
+# The OIE_JAVA_PATH has highest priority for specifying the Java executable.
+# If a valid version is found, the '-java-cmd' directive will be ignored when parsing the VM options file.
+if [[ -n "$OIE_JAVA_PATH" ]]; then
+    # Resolve path to an absolute one before validation
+    resolved_oie_java_path=$(resolve_absolute_path "$OIE_JAVA_PATH")
```
mgaffigan: Is this beneficial to resolve?  We're executing it just below (not passing it elsewhere).

server/basedir-includes/oieserver
tonygermano: It's useful if I have, for example, java 17 specified in the vmoptions file, but I want to specify the version from the command line without editing the file for testing.
```bash
$ OIE_JAVA_PATH=~/.sdkman/candidates/java/21.0.6.fx-zulu/bin/java ./oieserver
```

server/basedir-includes/oieserver
tonygermano: I should probably update the wording on that. It resolves to a cannonical path, so relative paths don't look like
```
/path/to/oie_home/../../other/path/to/bin/java
```
but instead like
```
/path/other/path/to/bin/java
```

server/basedir-includes/oieserver
mgaffigan: I think that would still work even if not normalized, and would disclose the `~` in logging.  I'll respond re: normalization in other thread.

server/basedir-includes/oieserver
mgaffigan: Is there a benefit to that?  I don't think it is a problem, just seems moderately complex thus hard to be confident does not introduce problems.  On the other hand, non-normalized paths definitely work.

To an extreme, I'd consider replacing the existing `eval` with `envsubst`.  I only used eval since it is not installed by default on Mac and is not included in busybox.  (Which, on balance, I still think `eval` is appropriate)

server/basedir-includes/oieserver
tonygermano: I went back and forth on that and decided to make all relative paths relative to OIE_HOME rather than the current vmoptions file. I don't know which is the "right" way to do it, but I did note in the comments both at the top of the script and in the resolve_absolute_path function where they are resolved from.

But now I realize that the vmoptions includes are relative to the current vmoptions file.

I asked Google Gemini which one I should change to match the other and it said to leave them alone
> This is an excellent and very subtle design question. You've correctly identified an inconsistency in how two different directives handle relative paths.
> 
> While it seems like they should be consistent, the current implementation where:
> 
> -include-options is relative to the current file
> -java-cmd is relative to $OIE_HOME
> ...is actually the ideal, most robust, and conventional design. You should not change either one.
> 
> Here’s the reasoning behind this seemingly inconsistent, but powerful, approach.
> 
> Two Different Use Cases
> The two directives serve fundamentally different purposes, and their path resolution logic is tailored to make each use case as easy and robust as possible.
> 
> 1. Use Case: Including Other Configuration Files (-include-options)
> Goal: To build a modular configuration by linking files together.
> 
> Why "Relative to Current File" is Correct: This allows you to create self-contained modules of configuration. You can create a directory like conf/ with a main.vmoptions that includes modules/database.vmoptions and modules/logging.vmoptions. This entire conf directory can be moved, copied, or symlinked to another location, and all its internal relative paths will remain intact. It makes the configuration portable and encapsulated.
> 
> If -include-options were relative to $OIE_HOME, then from within conf/main.vmoptions, you would have to write -include-options conf/modules/database.vmoptions. This is less intuitive and breaks encapsulation.
> 
> 2. Use Case: Pointing to an Application Resource (-java-cmd)
> Goal: To specify the location of a key application dependency, like a bundled Java runtime.
> 
> Why "Relative to $OIE_HOME" is Correct: The Java executable is a top-level resource for the entire application. No matter how deeply nested your include chain gets (e.g., conf/production/modules/tuning.vmoptions), you want a stable, predictable way to point back to the bundled JRE.
> 
> With the current design, the path is always -java-cmd jre/bin/java.
> 
> If -java-cmd were relative to the current file, a deeply nested config file would need to use a fragile and ugly path like -java-cmd ../../../jre/bin/java. This would break if you ever reorganized your configuration directories.

mgaffigan: Fair enough.  As long as it is not sensitive to cwd and documented, I think it is fine.

server/basedir-includes/oieserver
mgaffigan: Note: this is not a strong opinion.  As long as you have considered it, I have no issues with it.

server/basedir-includes/oieserver
tonygermano: I hear your points.

Right now it prints a warning to stderr that a specified java is incompatible before checking the next option on the priority list, so at least the troubleshooting part shouldn't be a problem. In this example, the `OIE_JAVA_PATH` failed because of a typo, and the `-java-cmd` failed because the version did not meet the minimum.
```bash
Warning: '/home/agermano/.sdkman/candidates/java/21.0.5.fx-zulu/bin/jaava' is specified by the OIE_JAVA_PATH environment variable, which is not a valid Java executable of at least version 17. Ignoring.
Warning: '/home/agermano/.sdkman/candidates/java/8.0.442.fx-zulu/bin/java' is specified by the -java-cmd directive in '/home/agermano/projects/connect/server/setup/conf/custom.vmoptions', which is not a valid Java executable of at least version 17. Ignoring.
Info: Found suitable java version specified by the JAVA_HOME environment variable
Starting Open Integration Engine...
Using Java from: /home/agermano/.sdkman/candidates/java/17.0.14.fx-zulu/bin/java
```

This is what the user sees when it exhausts all options
```
Warning: '/home/agermano/projects/connect/server/setup/~tony/.sdkman/candidates/java/21.0.5.fx-zulu/bin/java' is specified by the OIE_JAVA_PATH environment variable, which is not a valid Java executable of at least version 17. Ignoring.
Warning: '/home/agermano/.sdkman/candidates/java/8.0.442.fx-zulu/bin/java' is specified by the -java-cmd directive in '/home/agermano/projects/connect/server/setup/conf/custom.vmoptions', which is not a valid Java executable of at least version 17. Ignoring.
Warning: '/home/agermano/.sdkman/candidates/java/8.0.442.fx-zulu/bin/java' is specified by the JAVA_HOME environment variable, which is not a valid Java executable of at least version 17. Ignoring.
Warning: 'java' is specified by the system PATH, which is not a valid Java executable of at least version 17. Ignoring.
Error: Could not find a Java 17+ installation.
Please configure -java-cmd in conf/custom.vmoptions, set JAVA_HOME, set OIE_JAVA_PATH, or ensure 'java' in your PATH is version 17 or higher.
```

I could have it check `$OIE_HOME/jre/bin/java` after `vmoptions` and before `JAVA_HOME`?

server/basedir-includes/oieserver
tonygermano: I agree. I'll go with your second suggestion. It might be useful for someone to be able to copy/paste the entire command.

server/basedir-includes/oieserver
mgaffigan: I like checking for a jre dir. That makes packaging and replacement easy with a volume mount.

tonygermano: Then again, you could just put `-java-cmd jre/bin/java` in your `custom.vmoptions` and it would do the same thing (since it's relative to OIE_HOME,) and it wouldn't complain about that path being missing every time the script ran for people using JAVA_HOME or system PATH.

mgaffigan: I agree that putting it in custom.vmoptions produces the same result.  If it were added to the script separately, I would do so by defaulting `OIE_JAVA_PATH` to the bundled JRE if the variable is not set in the environment, but custom.vmoptions does seem cleaner.

I would still take `OIE_JAVA_PATH` or `-java-cmd` regardless of `is_valid_java_version`.  Those are both OIE specific requests to use a specific java, and ignoring them in favor of `JAVA_HOME` or `java` doesn't make sense to me.  If  `OIE_JAVA_PATH` or `-java-cmd` specified, they should be used.  If you are sure `is_valid_java_version` is accurate for all past, current, and future JRE's I would die returning failure.  Otherwise it should launch with the specified java, and maybe warn that it smells funny.

tonygermano: I'll think about how I might change this, but I can't both default `OIE_JAVA_PATH` to a bundled path and fail to start when `OIE_JAVA_PATH` doesn't point to a valid version, since the bundled version will not always be present.

tonygermano: There is also the possibility that the directive may be specified more than once in the same or (more likely) multiple vmoptions files, and it will take the last valid one, but it probably(?) shouldn't fail fast if the first one it encounters isn't valid.

mgaffigan: I think you are right in saying a packaged JVM can be specified in vmoptions. I don't see any relative benefit to detecting it separately. 

As for the same option specified multiple times, I would presume the last one is controlling, but that seems enough of a corner case as to leave it unspecified.  I would not imagine each to be checked for validity, and only valid options to be controlling.

server/basedir-includes/oieserver
tonygermano: So, it's the `eval` that expands the tilde, and that's already happening in the vmoptions files for the env variable expansion.  If I don't expand it, then I'll have to treat it like an absolute path or it will prepend $OIE_HOME.

server/basedir-includes/oieserver
tonygermano: I think the benefit is if you have a path that includes multiple `..` (either relative or one where they follow `${variables}`) and validation fails, that it makes it easier to see what it is resolving to in the error message in case a mistake was made. It's using the same method to obtain the canonical path as is used to obtain OIE_HOME in your original script. The AI said that is a safe and portable way to do it.

server/basedir-includes/oieserver
tonygermano: I pushed a commit with changes.
- OIE_JAVA_PATH will now fail-fast if not valid.
- `-java-cmd` validation no longer occurs while parsing. It will only validate the last directive encountered and fail-fast if not valid.
- JAVA_HOME only issues a warning if it is invalid so that it can still check the PATH.

I left the current method of checking the java version in place. It works for all known versions of java not as an official standard, but a conventional one. The "correct" way to get the "major" version is to actually include and attempt to run a java program that outputs `java.lang.Runtime.version().feature()`, but that was not supported until java 10. Prior to that, from within Java, you would use a system property that still required string parsing and was basically what I'm doing now. I think it's ok to leave it for now, and we can adjust it later if someone encounters a java version that doesn't work, but should. It is an easy workaround for them to modify the script to allow it to work until it gets fixed.

I do like having the script do the validity check, because the failure message is much more user friendly than if we just attempted to run it without checking first (and the fall-through checking would not be possible without it, though that is less important now.) 

The last change I made was prepend  `mirth-server-launcher.jar` to the beginning of the classpath if it sees that it was removed during vmoptions parsing.

pacmano1 (changes requested):
You need to fix the java check to handle 1.8.x (for a better error messagE)

tonygermano:
> I think you need to fix the java check to handle 1.8.x (for a better error message)

It won't change the error message, because it only reports that the given path does not meet the minimum requirement, not which version, if any, was found.

kayyagari (approved):
Tested on M2 with Java17.

pacmano1 (approved):
I can agree to disagree.   I think the version should be echoed when there's a problem, but will approve anyway.

kayyagari:
@pacmano1 the script reported this error when I first ran with JAVA_HOME set to 1.8:

`Warning: '/Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/bin/java' is specified by the JAVA_HOME environment variable, which is not a valid Java executable of at least version 17. Ignoring.
Warning: 'java' does not exist in your PATH or is not a valid Java executable of at least version 17.
Error: Could not find a Java 17+ installation.
Please configure -java-cmd in conf/custom.vmoptions, set OIE_JAVA_PATH, set JAVA_HOME, or ensure 'java' in your PATH is version 17 or higher.
`

Any other detail you would prefer in addition to this?.

pacmano1:
It's fine

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+#                   executable to be used. Can be a relative path from the
+#                   location of this script or a tilde path
+#                   (e.g., ~/path/to/java).
+#   JAVA_HOME       - Set the path to the root of a Java installation. The
+#                   script will look for 'bin/java' within this path.
+# =============================================================================
+
+# --- Script Parameters and Initial Variables ---
+param(
+    [string[]]$AppArgs = $args
```
mgaffigan: I think this would canonically be `[ValueFromRemainingArguments][string[]] $AppArgs`.  $args will work too, but doesn't show up in help.

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+# Set OieHome to the script directory using PowerShell's built-in variable
+$OieHome = $PSScriptRoot
+# The engine expects it's working directory to be OieHome
+try {
+    Set-Location -Path $OieHome -ErrorAction Stop
+}
+catch {
+    Write-Error "Could not change to the OieHome directory: $OieHome"
+    exit 1
+}
```
mgaffigan: I'm all for cautious code, but I can't imagine a case where the catch would be needed.  `Push-Location`  with a try/finally is often used to ensure the user's CWD does not change due to script execution.

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+        $PathToResolve = Join-Path -Path $OieHome -ChildPath $PathToResolve
+    }
+    
+    try {
+        return (Resolve-Path -LiteralPath $PathToResolve).Path
+    }
+    catch {
+        return $PathToResolve
+    }
+}
```
mgaffigan: ~ will get resolved on linux by `Resolve-Path`, and it's not really a thing on windows.  %USERPROFILE% would be the equivalent on windows, which would be resolved by `[Environment]::ExpandEnvironmentVariables(...)`

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    # Define the regex pattern to find ${...} variables.
+    $regex = '\$\{([a-zA-Z_][a-zA-Z0-9_]*)\}'
+
+    # Use the static Replace method, passing the input line, the regex, and our evaluator.
+    $expandedLine = [regex]::Replace($Line, $regex, $evaluator)
+    
+    return $expandedLine
+}
```
mgaffigan: I think this can be done with `[Environment]::ExpandEnviromentVariables($Line -replace '\$\{([a-zA-Z_][a-zA-Z0-9_]*)\}', '%$1%')`.  The %FOO% syntax works on both windows and linux in powershell.

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        $line = Expand-LineVariables -Line $line
+
+        if ($line -match '^-include-options\s+(.+)') {
+            $includedFile = $matches[1].Trim()
+            # Resolve relative paths to the current file's directory
+            if (-not [System.IO.Path]::IsPathRooted($includedFile)) {
+                $includedFile = Join-Path -Path (Split-Path -Path $File -Parent) -ChildPath $includedFile
+            }
```
mgaffigan, line 185: Join path already checks for rooted paths and can be done without check.

server/basedir-includes/oieserver.ps1
```diff
@@ -0,0 +1,293 @@
...
+}
+
+# Check -java-cmd from vmoptions (fail-fast on invalid, only if not already found).
+if (-not $FinalJavaCmd -and -not [string]::IsNullOrWhiteSpace($script:VmOptionsJavaCmd)) {
+    if (Test-IsValidJavaVersion -JavaCmd $script:VmOptionsJavaCmd) {
+        Write-Host "Info: Found suitable java version specified by the -java-cmd directive in '$($script:VmOptionsJavaCmdFile)'" -ForegroundColor Green
+        $FinalJavaCmd = $script:VmOptionsJavaCmd
+    } else {
+        Write-Error "'$($script:VmOptionsJavaCmd)' is specified by the -java-cmd directive in '$($script:VmOptionsJavaCmdFile)', which is not a valid Java executable of at least version $MinJavaVersion. Exiting."
+        exit 1
```
mgaffigan, line 240: Write-Error will automatically set a non-zero exit code.  When combined with `$ErrorActionPreference = "Stop"`, it will also stop script execution.

mgaffigan (changes requested):
See comments and suggested edits commit in https://github.com/mgaffigan/oie/commit/a6cd2efb084140f38d95da1696b878873b83b833.

Most are nits or concision.  Only significant issue I see in the powershell rev is that the args passed to java should be a splat (@ not $) - not an implicit array to string, since that will not handle quotes.

server/basedir-includes/oieserver.ps1
tonygermano: It's not something I would have thought to add myself, but AI added the check to the bash script when I told it to change to the directory at the start of the script, and I figured it couldn't hurt. Then it translated it to this for the powershell version, so I just kept it.

I can't really think of a need for Push-Location either, except maybe to use it during the recursive vmoptions parsing. Initially we just want to set the location to the script home, and never go back to whatever it was when the script was started.

server/basedir-includes/oieserver.ps1
tonygermano: I can probably take out the check for `$env:HOME` and only use `$env:USERPROFILE`, but in powershell on windows I can `cd ~` and it will change to my USERPROFILE dir. So, while traditionally it wasn't a thing on Windows, it is a thing in powershell, and it's a lot shorter to type. There also isn't any environment variable expansion happening in this function, but this function does run on the value of $OIE_JAVA_PATH.

server/basedir-includes/oieserver.ps1
tonygermano: nevermind that last comment. I see how you are using push-location and pop-location at the end

server/basedir-includes/oieserver.ps1
tonygermano: It wasn't my intention for this script to be run on non-Windows platforms, but if it is, then I should leave the $env:HOME check in.

server/basedir-includes/oieserver.ps1
tonygermano: This is for when the argument is an absolute path like "c:\path\to\my.vmoptions" and joining with the parent directory is not desired.

server/basedir-includes/oieserver.ps1
mgaffigan: Hm.  Ok. Didn't realize that powershell supported ~.  Regardless, it appears to be resolved by `Resolve-Path`, so I still don't think it should be done explicitly by the script.

Windows:
```
PS C:\> Resolve-Path ~\desktop

Path
----
C:\Users\mgaffigan\desktop
```

Ubuntu:
```
PS /> resolve-path ~/.bashrc

Path
----
/home/mgaffigan/.bashrc
```

server/basedir-includes/oieserver.ps1
mgaffigan: Got it.  You might use `[Path]::Combine(...)` then.

```
PS C:\drop> [System.IO.Path]::Combine('C:\drop\', 'c:\users\')
c:\users\
PS C:\drop> [System.IO.Path]::Combine('C:\drop\', 'a\')
C:\drop\a\
```

server/basedir-includes/oieserver.ps1
tonygermano: did not end up doing this because the desired behavior is to leave the original token in place when the variable is unset (I verified this is what install4j does)

server/basedir-includes/oieserver.ps1
tonygermano: I chose not to do this because it was working the way that I had it, and I think it's clearer to the reader. Probably could have gone either way.

server/basedir-includes/oieserver.ps1
tonygermano: I was reluctant to use resolve-path to do all of the expansion for the same reason that I removed `eval` from the bash script. I didn't want it doing too much (like wild-card expansion.) I also wanted it to be able to resolve a path whether the user gave `c:\path\to\java` or `c:\path\to\java.exe` on windows. It works the same way as the bash script and only resolves the parent directory, then checks if the parent+leaf is a command when testing the java version.

server/basedir-includes/oieserver.ps1
tonygermano: suggestion added

server/basedir-includes/oieserver.ps1
tonygermano: suggestion added

mgaffigan (approved):

kayyagari (approved):

jonbartels (approved):

gibson9583 (approved):

PR #123, opened 2025-06-21
Fix usernames being case sensitive

### LINKED ISSUE #122: [BUG] Username is case-sensitive when logging into MC Administrator
**Describe the bug**
Login to Mirth Connect Administrator fails if the case of the username doesn't match.

**To Reproduce**
Try to log in to MCA with different case on the letters.

**Expected behavior**
Login succeeds.

**Actual behavior**
Login fails with invalid username/password on Postgres, Derby, Oracle.
Login fails with "There was an error connecting to the server..." on SQL Server and MySQL.

**Related**
[Original bug #5569](https://github.com/nextgenhealthcare/connect/issues/5569)
Summary table in #121

mgaffigan:
Closes #122 / original [#5569](https://github.com/nextgenhealthcare/connect/issues/5569) by:

- Normalizing username after login success
- Using case-insensitive comparisons in DB

Implementation Notes:
- Used `LOWER()` in pgsql instead of collation to avoid having to rely upon system defaults or database defined collation.  Presumably table is not so large that sargability is a concern (since we're loading the full table to the client on each login)
- No collation support seems to exist in Derby, so `LOWER()`
- MSSQL and MySQL already case insensitive by system default

### DIFF
--- a/server/dbconf/derby/derby-user.xml
+++ b/server/dbconf/derby/derby-user.xml
@@ -33,7 +33,7 @@
 		FROM PERSON
 		<where>
          	<if test='id != null'>ID = #{id}</if>
-         	<if test='username != null'>AND USERNAME = #{username}</if>
+         	<if test='username != null'>AND LOWER(USERNAME) = LOWER(#{username})</if>
       	</where>
 	</select>
 
--- a/server/dbconf/oracle/oracle-user.xml
+++ b/server/dbconf/oracle/oracle-user.xml
@@ -33,7 +33,7 @@
 		FROM PERSON
 		<where>
 			<if test="id != null">ID = #{id}</if>
-			<if test="username != null">AND USERNAME = #{username}</if>
+			<if test="username != null">AND USERNAME COLLATE BINARY_CI = #{username}</if>
 		</where>
 	</select>
 
--- a/server/dbconf/postgres/postgres-user.xml
+++ b/server/dbconf/postgres/postgres-user.xml
@@ -29,7 +29,7 @@
 		FROM PERSON
 		<where>
          <if test='id != null'>AND ID = #{id}</if>
-         <if test='username != null'>AND USERNAME = #{username}</if>
+         <if test='username != null'>AND LOWER(USERNAME) = LOWER(#{username})</if>
       </where>
 	  </select>
    <select id='getUserCredentials'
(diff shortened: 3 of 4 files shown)


mgaffigan:
@david-labs-ca, there's discussion on the [original ticket](https://github.com/nextgenhealthcare/connect/issues/5569).  

I think the world at large usually assumes usernames to be case-insensitive.  It's also just plum broken on several case-insensitive database engines in main, so if the decision were to make them case sensitive, a change would still be required.

tonygermano:
suggestion: don't assume anything about which charset, collation, or locale are in use

I think it would be a good idea to use `LOWER(USERNAME) = LOWER(#{username})`  across all databases, and if that returns multiple results because a case-sensitive collation is in use and there are similar names, then fall back to `USERNAME = #{username}`.

tonygermano:
> @tonygermano Re: collation, at a certain level every string compare is done in the context of a collation, implicit or explicit. `UPPER` and `LOWER` also produce sargability challenges. In my experience, you will make more DBA friends with `COLLATE` than with `LOWER(...)`. The only reason I did not use `COLLATE` uniformly is to avoid DDL changes.

@mgaffigan  I don't think the PERSON table is expected to get very large where a scan would be a problem, and this seems evident because there isn't an index on USERNAME. In this case, I think it's better to be safe and use the standard SQL function without making assumptions about the environment rather than going for performance. LOWER should work across all databases respecting their current collation, character encoding, and locale.

This is completely up to you, but if you want to address this issue, too, it's related https://github.com/nextgenhealthcare/connect/issues/3386

mgaffigan:
> @mgaffigan for completeness, what are your thoughts on updating `mysql-user.xml` and `sqlserver-user.xml`, too?

If it will get the PR merged, I'm unopposed.

I do not think it is called for, but I do not like `lower` to start with. If those default collations are changed, the system is likely running with locale specific collation that is more correct than what `lower` would provide. Running with a non-default collation is also something I would call an unsupported configuration.

From a maintenance perspective I think it is counterproductive to keep the different database server scripts in sync: I've never found SQL to be an actually portable language at scale (DDL aside, the query engines are too different). The appropriate interface for changing persistence is at the DAL - not at SQL.  This also allows in memory stores for unit tests, no-SQL, etc.

tonygermano (approved):

gibson9583:
> I don't know how this will affect the user roles plugin from NG - maybe we don't care?
> 
> 

We don't care. This will release in a version number unsupported by any open source NG release. With their plugin architecture change - bwc won't work either.

mgaffigan:
> I don't know how this will affect the user roles plugin from NG

@pacmano1, I can't see how it would be affected. No API changes in this PR, and the case sensitive username is still used after login.

PR #125, opened 2025-06-21
Avoid crash when invalid URI is passed to test file write

### LINKED ISSUE #124: [BUG] Mirth Client Crashes when doing a Test Write on File Writer with no value in Directory field
**Describe the bug**
Mirth Client Crashes when doing a Test Write on File Writer with no value in Directory field and Mirth 3.12 running inside of a container.

**To Reproduce**
Create a channel and set destination to File Writer
Click Test Write

**Expected behavior**
Not crash and given an error.

**Actual behavior**
An unexpected error occurred.  [Full message](https://github.com/user-attachments/files/20848301/writefiletest.txt)

**Related**
Originally [#4963](https://github.com/nextgenhealthcare/connect/issues/4963)
Summary #121

mgaffigan:
Closes #124 / original [#4963](https://github.com/nextgenhealthcare/connect/issues/4963).

Updates server to provide a normal failure when a URI can't be parsed.

Implementation notes:
- I did not add separate client-side verification since it's unclear to me how to validate in a manner that supports all of the various file write options.  The new error when no directory is specified is 
new error

### DIFF
--- a/server/src/com/mirth/connect/connectors/file/FileConnectorServlet.java
+++ b/server/src/com/mirth/connect/connectors/file/FileConnectorServlet.java
@@ -65,7 +65,12 @@ public class FileConnectorServlet extends MirthServlet implements FileConnectorS
             int timeout = Integer.parseInt(timeoutString);
 
             FileConnector fileConnector = new FileConnector(channelId, connectorProperties, null);
-            URI address = fileConnector.getEndpointURI(host, scheme, schemeProperties, secure);
+            URI address;
+            try {
+                address = fileConnector.getEndpointURI(host, scheme, schemeProperties, secure);
+            } catch (Exception e) {
+                return new ConnectionTestResponse(ConnectionTestResponse.Type.FAILURE, "Unable to parse URL, Reason: " + e.getMessage());
+            }
             String addressHost = address.getHost();
             int port = address.getPort();
             String dir = address.getPath();


kryskool (approved):

server/src/com/mirth/connect/connectors/file/FileConnectorServlet.java
```diff
@@ -65,7 +65,12 @@ protected ConnectionTestResponse testReadOrWrite(String channelId, String channe
             int timeout = Integer.parseInt(timeoutString);
 
             FileConnector fileConnector = new FileConnector(channelId, connectorProperties, null);
-            URI address = fileConnector.getEndpointURI(host, scheme, schemeProperties, secure);
+            URI address;
+            try {
+                address = fileConnector.getEndpointURI(host, scheme, schemeProperties, secure);
+            } catch (Exception e) {
```
jonbartels, line 71: Can/should specific exception types be caught or is a pokemon handler (gotta catch em all!)  OK here?

tonygermano: I was looking at this one last night. I think it's fine if we don't want to leak uncaught exceptions. There's only one line in the try, so we know what failed, and the exception message is included.

mgaffigan: We can drop down to just catching `URISyntaxException`, but I'm not sure it is any worse/better.  The catch at line 103 will swallow everything regardless.

tonygermano: My vote is to keep it catching all Exceptions.

kpalang (approved):

server/src/com/mirth/connect/connectors/file/FileConnectorServlet.java
jonbartels: I'm OK with catching all exceptions. Ty for discussing!

PR #126, opened 2025-06-25
#40 Add initial Dockerfile

kpalang:


docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
+# syntax=docker/dockerfile:1
+
+ARG ALPINE_JRE_TAG=17.0.15_6-jre-alpine
+ARG ALPINE_JDK_TAG=17.0.15_6-jdk-alpine
+ARG UBUNTU_JRE_TAG=17.0.15_6-jre-noble
+ARG UBUNTU_JDK_TAG=17.0.15_6-jdk-noble
```
jonbartels, line 6: Each of these is ALMOST the same thing. 

What about defining "JAVA_VERSION = 17.0.15" then bulding the tags from that so that we have ONE place to update the java version?

Alpine also supports just "17" or "21" do we need the 17.0.5 full notation?

docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+
+ARG UID=14285
+ARG GID=14285
```
jonbartels, line 9: Is 14285 significant or just a random number?

docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+
+WORKDIR /opt
+
+# Download Open Integration Engine release
+RUN apk add --no-cache curl \
+    && curl -L \
+      -o /opt/engine.tar.gz \
+      -H "Accept: application/vnd.github+json" \
+      -H "X-GitHub-Api-Version: 2022-11-28" \
+      https://github.com/OpenIntegrationEngine/engine/releases/download/v4.5.2-tp.1/oie_unix_4_5_2.tar.gz \
```
jonbartels, line 24: Should this be an ARG like the nextgen image?

docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+
+USER engine
+ENTRYPOINT ["./entrypoint"]
+CMD ["./oieserver"]
+
+##########################################
+#
+#     Ubuntu Images
+#
+##########################################
```
jonbartels, line 89: Why do we want both alpine and ubuntu images?

docker/entrypoint.sh
```diff
@@ -0,0 +1,244 @@
+#!/bin/bash
```
jonbartels: Does this need an MPL header or copyright since its clearly yanked from NextGen?

docker/Dockerfile
kpalang: Random. It is to try to eliminate conflicts with existing users.

docker/Dockerfile
kpalang: I will be. I've yet to figure out how to accommodate both the Github "Release" object and downloading arbitrary artifacts from actions.

docker/Dockerfile
kpalang: https://www.youtube.com/watch?v=QQ9ARfHCwOY

docker/entrypoint.sh
kpalang: Probably. Will investigate.

docker/Dockerfile
kpalang: I think here specifying full tags is better, as realistically, there's not a single standard they're using for tags. So generating them on the fly introduces the risk of finding a non-existing tag.

docker/entrypoint.sh
tonygermano: Yes. I've been putting SPDX headers on all new files.

tonygermano:
```
#!/usr/bin/env bash
#
# SPDX-License-Identifier: MPL-2.0
# SPDX-FileCopyrightText: 2023 NextGen Healthcare
#
```

docker/entrypoint.sh
```diff
@@ -0,0 +1,249 @@
...
+		curl -sSLf "${KEYSTORE_DOWNLOAD}" -o "$APP_DIR/appdata/keystore.jks" || echo "problem with keystore download"
+	fi
+fi
+
+# if delay is set as an environment variable then wait that long in seconds
+if ! [ -z "${DELAY+x}" ]; then
+	sleep $DELAY
+fi
+
+exec "$@"
```
tonygermano, line 249: missing newline at end of file

docker/build-images.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Build alpine-jre image
+docker build -f Dockerfile --target alpine-jre -t openintegrationengine/engine:latest-alpine-jre .
+# Build ubuntu-jre image
+docker build -f Dockerfile --target ubuntu-jre -t openintegrationengine/engine:latest-ubuntu-jre .
+
+# Build alpine-jdk image
+docker build -f Dockerfile --target alpine-jdk -t openintegrationengine/engine:latest-alpine-jdk .
+# Build ubuntu-jdk image
+docker build -f Dockerfile --target ubuntu-jdk -t openintegrationengine/engine:latest-alpine-jdk .
```
tonygermano, line 13: missing newline at end of file

docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+    && mv /opt/oie /opt/engine \
+    && mkdir -p /opt/engine/appdata
+
+WORKDIR /opt/engine
+COPY --chmod=755 ./entrypoint.sh /opt/engine/entrypoint
+
+RUN rm -rf cli-lib manager-lib \
+    && rm mirth-cli-launcher.jar oiecommand
+
+RUN (cat oieserver.vmoptions docs/oieservice-java9+.vmoptions ; echo "") > oieserver_base.vmoptions
```
tonygermano, line 35: this step shouldn't be necessary. Our default vmoptions already include the java 9+ options

docker/Dockerfile
tonygermano: I need to check some of the nextgen issues, but I think it is desirable to conflict with existing users when doing bind mounts.

docker/Dockerfile
jonbartels: There is a clear standard. The common images use this pattern:

```ARG JAVA_VERSION=17
ARG ALPINE_JRE_TAG=${JAVA_VERSION}-jre-alpine
ARG ALPINE_JDK_TAG=${JAVA_VERSION}-jdk-alpine
ARG UBUNTU_JRE_TAG=${JAVA_VERSION}-jre-noble
ARG UBUNTU_JDK_TAG=${JAVA_VERSION}-jdk-noble
```

Thus you get the latest 17.whatever.something_patch

docker/build-images.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+# Build alpine-jre image
+docker build -f Dockerfile --target alpine-jre -t openintegrationengine/engine:latest-alpine-jre .
```
jonbartels, line 6: I sent you a patch on Discord to DRY this up and to support arm64 architectures.

jonbartels:
I have a local commit of `docker/README.md` . Can you edit this PR to allow others to push to it so I can contribute?

I have launched an image. OIE starts and runs. I can log in. I can create and deploy channels. 

There is a logged error:
```ERROR 2025-06-26 01:38:35.473 [Main Server Thread] Server: Thu Jun 26 01:38:35 UTC 2025 Thread[Main Server Thread,5,main] java.io.FileNotFoundException: derby.log (Permission denied)```

I thing that there just needs to be a persistent volume maybe? The error does not appear to be fatal, I'll dig into it.

File: docker/Dockerfile (hunk 1 above)
Comment (tonygermano) [reply docker/Dockerfile:6]: We should use the full version in the tag rather than just a major version. That's good practice in docker just like it is with software library versions. I think there's a way with dependabot or other bots to have it automatically open PRs to upgrade the version when new ones are released.

**Comment by mgaffigan** [comment]
Here are the examples that were pulled out of the old repo: https://github.com/OpenIntegrationEngine/engine/commit/0d8227670299f6c9fef43aeda2617582b679e495.  Not sure if it is easier to pull from what I had to start over from the original repo.

Includes:
- Samples with rebranding to OIE
- Suggested edits to main readme for docker quick-start using production image

Does not include:
- [Tests](https://github.com/nextgenhealthcare/connect-docker/tree/master/test) (couldn't get them to run)

File: docker/Dockerfile [hunk 10]
```diff
@@ -0,0 +1,129 @@
...
+      -o /opt/engine.tar.gz \
+      -H "Accept: application/vnd.github+json" \
+      -H "X-GitHub-Api-Version: 2022-11-28" \
+      https://github.com/OpenIntegrationEngine/engine/releases/download/v4.5.2-tp.1/oie_unix_4_5_2.tar.gz \
+    && tar xzf engine.tar.gz \
+    && mv /opt/oie /opt/engine \
+    && mkdir -p /opt/engine/appdata
+
+WORKDIR /opt/engine
+COPY --chmod=755 ./entrypoint.sh /opt/engine/entrypoint
```
mgaffigan, line 30: I suggest either the checked-in file should not have the .sh extension or the copy should not change the name and `ENTRYPOINT ["./entrypoint.sh"]`.  This seems to add unnecessary differences between dev and release file structure.

docker/entrypoint.sh
mgaffigan, line 1: Might this go in server/base-includes named `configure-from-env.sh` or something?  Seems generally useful and not tied to docker.  APP_DIR can just be the directory of the script.

docker/entrypoint.sh
```diff
@@ -0,0 +1,249 @@
...
+			else # otherwise key exists, overwrite it
+				# echo "key ${ACTUAL_KEY} exists, overwriting. Value = ${VALUE}"
+				ESCAPED_KEY="${ACTUAL_KEY//./\\.}"
+				sed -i "s/^${ESCAPED_KEY}\s*=\s*.*\$/${ACTUAL_KEY} = ${VALUE//\//\\/}/" "$APP_DIR/conf/mirth.properties"
+			fi
+		fi
+	fi
+done <<< "`printenv`"
+
+cp oieserver_base.vmoptions oieserver.vmoptions
```
mgaffigan, line 134: Can this not be done from the dockerfile?  This breaks use if oieserver_base.vmoptions does not exist.

docker/Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+
+VOLUME /opt/engine/appdata
+VOLUME /opt/engine/custom-extensions
+WORKDIR /opt/engine
+
+EXPOSE 8443
+
+USER engine
+ENTRYPOINT ["./entrypoint"]
+CMD ["./oieserver"]
```
mgaffigan, line 83: Is there a reason to duplicate for JRE and JDK?  They are the same, and since the base tag is already an argument, it can be passed from build-images.sh.

I understand having two targets for alpine vs ubuntu, since those have differences.

mgaffigan (changes requested):

docker/entrypoint.sh
tonygermano: base_includes goes into the root project directory of the final release, and this script probably should not be included in the release

docker/entrypoint.sh
tonygermano: There shouldn't be a reason to do anything with the vmoptions. They are included in our source repo, but they were not in the nextgen one.

docker/entrypoint.sh
```diff
@@ -0,0 +1,249 @@
...
+
+# Address reflective access by Jackson
+echo "--add-opens=java.desktop/java.awt.color=ALL-UNNAMED" >> oieserver.vmoptions
```
tonygermano, line 137: Also not necessary. It's already included

https://github.com/OpenIntegrationEngine/engine/blob/8e6a11ef6898eb5ffca3d2606183027fb2f7f10d/server/conf/default_modules.vmoptions#L20

docker/entrypoint.sh
mgaffigan: I understand that, but the script does not seem overly related to docker.  It seems like it could be useful in other circumstances to generate configuration from environment variables.

I have no objection to keeping it as a docker-only thing, but I think it is worth considering.

kpalang:
Closing this as it is not ready and should not receive feedback yet.

Issue #121, opened 2025-06-20
[IDEA] Review priority bugs from Mirth Connect


rogin:
Similar to what Innovar [performed](https://github.com/Innovar-Healthcare/BridgeLink/issues/19) which I liked, I pulled what I considered worthwhile from Mirth Connect's bug label, which also includes a few features and a security option worth investigating. These were copied to a local Excel file about a month ago, so a few more may be added should I recheck. Initial sizing and proposed value also provided.

Helpful Excel to Markdown converter [here](https://tabletomarkdown.com/convert-spreadsheet-to-markdown/).

Excel file for convenience: [OIE_task_migration.xlsx](https://github.com/user-attachments/files/20843344/OIE_task_migration.xlsx)

| OIE Issue or PR | General Issue                                                                                                    | NextGen Issue #                                                                                                      | NextGen Pull #                                                                                                   | Type     | T-Shirt Size | Proposed Value | Comment                                                                   |
| ---- | ---------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------- | -------- | ------------ | -------------- | ------------------------------------------------------------------------- |
|| Mirth Connect 4.1.1.b303 creates corrupted keystore.jks when using "encryption.properties = 1"                   | [5466](https://github.com/nextgenhealthcare/connect/issues/5466) | NONE                                                                                                             | BUG      |              | High           |                                                                           |
|| DB Reader... (Generate).. Select Button..Get Tables... empty string does NOT retrieve all tables                 | [4876](https://github.com/nextgenhealthcare/connect/issues/4876) | NONE                                                                                                             | BUG      |              | Medium         |                                                                           |
|| Attachment Handler does not accomodate more than 40 characters for mime type                                     | [4479](https://github.com/nextgenhealthcare/connect/issues/4479) | NONE                                                                                                             | BUG      |              |                |                                                                           |
|| Client GlobalMap Dashboard Viewer does not correctly display multiple entries where the keys only differ by case | [5314](https://github.com/nextgenhealthcare/connect/issues/5314) | NONE                                                                                                             | BUG      |              |                |                                                                           |
|| DefaultConfigurationController.initializeSecuritySettings writes the keystore unnecessarily                      | [5467](https://github.com/nextgenhealthcare/connect/issues/5467) | NONE                                                                                                             | BUG      |              | High           | Marked high value as it would preserve file's timestamp                   |
[...]

PR #129, opened 2025-07-02
Updates Migrate4_5_2.java to allow for log4j2-cli.properties not being present

rogin:
Adds null check as property file may not exist.

To Reproduce: upgrade 4.5.1 to 4.5.2 without log4j2-cli.properties in the /conf directory

Related
https://github.com/nextgenhealthcare/connect/pull/6327


tonygermano (approved):
I updated this PR. The original request in the NextGen repo stated that the mirth-cli had been uninstalled from the server install directory, and so the log4j2-cli.properties file was not present.

Technically, even the log4j2.properties file is not strictly required, and migration should not fail if it is missing.

This has been refactored so each file is processed in the same way and migration will progress if either is missing, logging an INFO level message when a file can't be found.

Additionally, it will now only attempt to save the file if it actually made a change to it.

PR #130, opened 2025-07-03
Default 24 hour notation for search criteria

rogin:
Adds a checkbox to the Messages Browser and Events Browser which toggles the time format of the time input fields between `HH:mm` and `hh:mm aa`. On form load and on search reset, it defaults to 24hr notation.

I used most of the previous PR, but I changed the listener to one that responds to `JCheckbox.setSelected(bool)`. Changes to the `.form` files were also included, but those may be completely unused.

**Related**
Previous PR - https://github.com/nextgenhealthcare/connect/pull/5989

### DIFF
--- a/client/src/com/mirth/connect/client/ui/browsers/event/EventBrowser.form
+++ b/client/src/com/mirth/connect/client/ui/browsers/event/EventBrowser.form
@@ -55,6 +55,7 @@
               <Group type="103" groupAlignment="0" attributes="0">
                   <Group type="102" attributes="0">
                       <Component id="allDayCheckBox" min="-2" max="-2" attributes="0"/>
+                      <Component id="hourNotation24" min="-2" max="-2" attributes="0"/>
                       <EmptySpace min="-2" pref="19" max="-2" attributes="0"/>
                       <Group type="103" groupAlignment="0" max="-2" attributes="0">
                           <Component id="levelBoxInformation" alignment="0" pref="95" max="32767" attributes="0"/>
@@ -163,7 +164,10 @@
                           <Component id="allDayCheckBox" alignment="3" min="-2" max="-2" attributes="0"/>
                       </Group>
                       <EmptySpace max="-2" attributes="0"/>
-                      <Component id="levelBoxWarning" min="-2" pref="16" max="-2" attributes="0"/>
+                      <Group type="103" groupAlignment="3" attributes="0">
+                          <Component id="levelBoxWarning" min="-2" pref="16" max="-2" attributes="0"/>
+                          <Component id="hourNotation24" alignment="3" min="-2" max="-2" attributes="0"/>
+                      </Group>
                       <EmptySpace max="-2" attributes="0"/>
                       <Component id="levelBoxError" pref="16" max="-2" attributes="0"/>
                   </Group>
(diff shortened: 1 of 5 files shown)


mgaffigan:
Can this respect the OS preference instead?  It looks like Java [provides access](https://www.oracle.com/java/technologies/javase/jdk11-suported-locales.html#providers:~:text=HOST%3A%20A%20provider,and%20macOS%20platform.) to the user's preferred 12/24 time format via `java.locale.providers=HOST` for Windows and Mac on Java 11+.

```java
import java.time.LocalTime;
import java.time.format.DateTimeFormatter;
import java.time.format.FormatStyle;

public class Main {
    public static void main(String[] args) {
        System.setProperty("java.locale.providers", "HOST");
        System.out.println(LocalTime.now().format(DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT)));
    }
}
```

Prints `17:44` on Windows with 24-hour time set in Control Panel on Java 17.

rogin:
> Can this respect the OS preference instead? It looks like Java [provides access](https://www.oracle.com/java/technologies/javase/jdk11-suported-locales.html#providers:~:text=HOST%3A%20A%20provider,and%20macOS%20platform.) to the user's preferred 12/24 time format via `java.locale.providers=HOST` for Windows and Mac on Java 11+.
> 
> ```java
> import java.time.LocalTime;
> import java.time.format.DateTimeFormatter;
> import java.time.format.FormatStyle;
> 
> public class Main {
>     public static void main(String[] args) {
>         System.setProperty("java.locale.providers", "HOST");
>         System.out.println(LocalTime.now().format(DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT)));
>     }
> }
> ```
> 
> Prints `17:44` on Windows with 24-hour time set in Control Panel on Java 17.

I like the idea. Your link omits linux compatibility in Oracle's v11, so perhaps linux guys can check their systems. If good, I could add a new slot in Settings>Administrator>User Preferences, with an option like "Use 24 hour notation for search criteria" with Yes, No, Use system default. I'd model the logic of reading and storing the boolean like used for "Import code template libraries with channels".

rogin:
Not bad
add-to-settings

tonygermano:
I'll try to bullet-point my previous comment. I'm not disagreeing with you about respecting user preferences.

1. The system property `java.locale.providers` is a user preference. The proposed change would ignore the user setting that property and override it.
2. I am unsure of the consequences of changing the default value from `CLDR,COMPAT` to `HOST` and I think that needs to be evaluated outside the context of this PR only.
3. 24-hour should likely be the **default** regardless of the setting for the user's system clock because
   1. The search results are already in 24-hour format, and that is not configurable. The search parameter should match.
   2. 24-hour format is easier to type
   3. Many people don't change their system clock to 24-hour format, nor do they have a desire to. Searching a log and glancing at a clock are not the same use case.
4. By storing it as a local preference, if the user sets the value to 12-hour one time, it should apply to all servers they access from that client device.

When it comes to the release notes I think this would capture the two choices we are discussing:
1. When searching messages or events, the time selection now defaults to 24-hour format. If you prefer to retain the previous behavior go to Settings-> User Preference and change the "Message searches use 24-hour format" value to "No."
2. When searching messages or events, the time selection now supports 24-hour format and will match the settings of your system clock by default. If you prefer to change this behavior to always use 12-hour or 24-hour format, go to Settings-> User Preference and change the "Message searches use 24-hour format" value to "Yes" or "No" accordingly.

I think the first option is cleaner and will require fewer people to actually need to go in and change the setting.

rogin:
I agree with Tony's first option. I'll update the code.

kryskool:
Hi 

In France we use the 24H date formatting, and when i go to View Message, i see the date in column is in the 24h format
and on the earch part, we have the 12h format, the information is therefore inconsistent

mstsc_rkQg6hQ4XH

I think the date column is the result come from the database, that's why we have the inforamtion in the 24h format

Regards,

PR #132, opened 2025-07-03
Updated for Java Version and external readme.html for 4.5.2

pacmano1:
Move html file outside of install4j and configure relative path name. 

Fixed bug with max java version.


tools/install4j/oie-installer-config.install4j
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <install4j version="11.0.3" transformSequenceNumber="11">
-  <directoryPresets config="/Users/dbugger/projects/governance/branding/icons" />
-  <application name="Open Integration Engine" applicationId="4145-9206-7630-8076" mediaDir="${compiler:installer:mediaRoot}/server/build" shortName="oie" publisher="Open Integration Engine Project" publisherWeb="https://openintegrationengine.com" version="4.5.2" backupOnSave="true" autoSave="true" macVolumeId="88477e584eb462ba" javaMinVersion="17">
+  <directoryPresets config="/Users/pcoyne/Documents/GitHub/engine/tools/install4j" />
```
gibson9583: what does this preset do and is it required? If not required, probably worth removing it just to remove your identity from the source

pacmano1: good catch - let me see if i can make that relative.

pacmano1: Line not needed so removed.

tools/install4j/oie-installer-config.install4j
```diff
@@ -827,9 +835,7 @@ return console.askYesNo(message, true);
       <jreBundle jreBundleSource="none" />
     </windowsArchive>
   </mediaSets>
-  <buildIds>
-    <mediaSet refId="43" />
-    <mediaSet refId="151" />
-    <mediaSet refId="153" />
+  <buildIds buildAll="false">
```
kayyagari, line 838: This flag requires to be set to true to while building release artifacts. It is fine to be left as is for now, but just wanted to bring it to attention.

kayyagari (approved):

tools/install4j/readme.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+<body style="font-family: Arial, sans-serif; line-height: 1.6; margin: 1em; background-color: white; color: black;">
+  <h1 style="color: black;">Open Integration Engine - 4.5.2 Release</h1>
+  <p>Open Integration Engine 4.5.2</p>
+  <p>
+    This release is recommended for users already familiar with Mirth Connect.
```
tonygermano: question: Is this statement appropriate?

Since this is a general release, should we remove this first sentence stating the release is recommended for users already familiar with Mirth Connect?

pacmano1: The intention there was a precursor for not asking for ports and so in the installer.

tonygermano: I think I would just drop that first sentence. The goal of the current installer is supposed to be simplicity, and it installs with sensible defaults. I think that might be discouraging to new users who read it when we are not offering an alternative release for users who are not already familiar with mirth.

pacmano1: This OK?

```
This installer sets up a fully functional version of the engine with minimal user input.
The installer requires administrator permissions to run and install services.
```

tonygermano: Yeah, I think that's good.

jonbartels (approved):

PR #134, opened 2025-07-04
Avoid writing the keystore if no changes were made

rogin:
Allow for using the keystore on a readonly filesystem. Currently, a write to the keystore will always occur. This change limits writes to only when necessary.

I welcome others to test by mounting the keystore as readonly. The keystore should not have the default storepass and keypass that that will trigger a write.

I'm considering a test in `DefaultConfigurationControllerTest` where the modification time of a keystore is checked before and after the call, but that's a lot of investigation to get it written correctly.

Minor other code cleanup was also included.

**Related**
Original ticket - https://github.com/nextgenhealthcare/connect/issues/5467
Previous PR by @jonbartels  - https://github.com/nextgenhealthcare/connect/compare/development...jonbartels:connect:5467-keytore-avoid-writes

### DIFF
--- a/server/src/main/java/com/mirth/connect/server/controllers/DefaultConfigurationController.java
+++ b/server/src/main/java/com/mirth/connect/server/controllers/DefaultConfigurationController.java
@@ -53,7 +53,6 @@ import javax.crypto.SecretKey;
 import javax.net.ssl.SSLSocketFactory;
 import javax.xml.parsers.DocumentBuilderFactory;
 
-import com.mirth.connect.client.core.BrandingConstants;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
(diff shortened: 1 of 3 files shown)


tonygermano (changes requested):
I haven't tested it yet, but I'm fine with this approach.

I would like to see the deprecation of ResourceUtil.closeResourceQuietly broken out to a different PR. It's not related to this keystore change, and the current state of the PR leaves a lot of references calling the deprecated method. I wonder if in some situations we would do better to change the calls to a try-with-resource block rather than using IOUtils.

Could you also please update the javadoc for the two private methods you changed to indicate the purpose of the return value?

jbeckers (approved):

jbeckers:
This bites me when rolling out our mirth instances with a custom keystore. Ansible sees the written-to keystore as a change and always overwrites it with ours.

PR looks good to me, though I agree the deprecation should be a different PR

jonbartels:
I have added tests to address copilot remarks.

I still need to build and manually test this locally. It should still be fine since the code didn't change in the last 2 years but I'll do that work and share my results.

I want to leave in the deprecation. It was observed as an issue while working this case. Leaving the deprecation flag helps others avoid using sub-optimal methods. Making it a separate PR adds to the growing pile of PRs and begging for reviews.

mgaffigan (approved):

jonbartels:
I have tested this feature. I am attaching the test docker-compose file here. This exercises the code by having a first boot write the store then the second boot reads it from a read-only mount. 

[docker-compose.yml](https://github.com/user-attachments/files/32068229/docker-compose.yml)

gibson9583 (approved):
LGTM - but I dont think @tonygermano items were addressed.

PR #139, opened 2025-07-11
Skip JAR permissions for unsigned builds

### LINKED ISSUE #138: [IDEA] Skip modifying permissions when disableSigning
Ant build specifying `-DdisableSigning=true` still modifies jar manifests (presumably in a way that only matters for signed jars being evaluated by the now EOL JWS).  I'm not sure any launchers are even evaluating permissions, given that the Java machinery to enforce them is dead.

Skipping this step for unsigned builds reduces build time by 55 seconds on my machine.  (66% reduction in build time, ignoring tests).

mgaffigan:
Disables the build step to add MANIFEST.MF permission attributes if signing is disabled.  Closes #138

### DIFF
--- a/server/build.xml
+++ b/server/build.xml
@@ -1051,22 +1051,6 @@
 		
 		<taskdef resource="net/sf/antcontrib/antlib.xml" classpathref="classpath" />
 
-		<!-- modify jar manifests -->
-		<!-- don't modify bcprov, since it already has the manifest changes and is signed by BC -->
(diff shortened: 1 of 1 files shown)


tonygermano (approved):
I added some additional context on the related issue. I think this is a smart change.

tonygermano (commented):
suggestion: expand the commit message

Add some detail to the message body about why this code exists and why we think it's ok to disable it now.

I have also been trying to add trailers to the commit messages like
```
Issue: https://github.com/OpenIntegrationEngine/engine/issues/138
```

Github will still recognize that and link the commit to the issue without needing to try to figure out how to wedge it into the commit subject. It's more easily discoverable by analysis tools, as well.

kpalang (approved):

PR #142, opened 2025-07-12
Remove empty extensionmanager-client.jar

tonygermano:
ExtensionManager apparently used to be a plugin, but it has not been for some time. The step in the client build script that produced the extension jar was not removed at the same time and has been producing an effectively empty jar file with no classes and only a manifest. The actual extensionmanager related classes are included in mirth-client.jar.

Related: 027ae2405ee6fe66b26b29a28e5315fcaf1dace0

### DIFF
--- a/client/ant-build.xml
+++ b/client/ant-build.xml
@@ -46,7 +46,6 @@
 		<property name="plugins.dashboardstatus" value="${extensions}/dashboardstatus" />
 		<property name="plugins.destinationsetfilter" value="${extensions}/destinationsetfilter" />
 		<property name="plugins.dicomviewer" value="${extensions}/dicomviewer" />
-		<property name="plugins.extensionmanager" value="${extensions}/extensionmanager" />
 		<property name="plugins.httpauth" value="${extensions}/httpauth" />
 		<property name="plugins.imageviewer" value="${extensions}/imageviewer" />
 		<property name="plugins.javascriptrule" value="${extensions}/javascriptrule" />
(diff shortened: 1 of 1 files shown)


mgaffigan (approved):

pacmano1 (approved):

tonygermano:
> That's not needed for the extensions view? Gets referenced in places like: `client/src/com/mirth/connect/client/ui/Frame.java`

See my comment in the description that the classes are included in mirth-client.jar instead, and the related commit. The classes are in a different package than where this build step is looking.

tonygermano:
> That's not needed for the extensions view? Gets referenced in places like: `client/src/com/mirth/connect/client/ui/Frame.java`

This is run from a nextgen mirth 4.5.0 docker container. You can see there are no class files in it.
```bash
mirth@f874da704e65:/opt/connect$ unzip -l extensions/extensionmanager/extensionmanager-client.jar 
Archive:  extensions/extensionmanager/extensionmanager-client.jar
  Length      Date    Time    Name
---------  ---------- -----   ----
      180  2024-01-17 12:29   META-INF/MANIFEST.MF
      231  2024-01-17 12:30   META-INF/SERVER.SF
     8994  2024-01-17 12:30   META-INF/SERVER.RSA
        0  2024-01-17 12:28   META-INF/
---------                     -------
     9405                     4 files
```

gibson9583 (approved):

kpalang (approved):

Issue #18, opened 2025-04-04
Target java 17 as the minimum version


tonygermano:
Stub to fill in later.

pmtorresv:
I agree with you @tonygermano OIE should keep the pace of Java LTS versions for the sake of performance and security.

Even two LTS versions ago would be fine if they are still under extended support as the support schedule is far from regular https://www.oracle.com/java/technologies/java-se-support-roadmap.html

PR #146, opened 2025-07-23
Update Dependencies from Bridgelink via Cherry Pick

gibson9583:
This seems like the most efficient way to handle updating these jars. Pulling from a known quantity rather than reviewing ourselves


gibson9583:
@mgaffigan   - thanks. Also pushed a commit to revert the change to DefaultMetaData

server/build.xml
```diff
@@ -1002,11 +1002,6 @@
 			<fileset dir="${docs}" />
 		</copy>
 
-		<!-- copy basedir files -->
-		<copy todir="${setup}">
-			<fileset dir="${basedir_includes}" />
-		</copy>
```
mgaffigan, line 1008: Not sure why this would be removed without oieserver.vmoptions removed.  Presumably bl is handling vmoptions differently.

server/build.xml
```diff
@@ -1048,99 +1043,41 @@
...
+			<sequential>
+				<retry retrycount="5" retrydelay="1000">
+					<signjar jar="@{jarFile}" alias="${key.alias}" keystore="${key.keystore}" storepass="${key.storepass}" keypass="${key.keypass}" storetype="${key.storetype}" tsaurl="http://timestamp.digicert.com" digestalg="SHA-256">
+						<!-- http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=7127374 -->
+						<sysproperty key="jsse.enableSNIExtension" value="false" />
+						<sysproperty key="https.protocols" value="TLSv1.2,TLSv1.1" />
+					</signjar>
+				</retry>
+			</sequential>
+		</for>
```
mgaffigan: I think all of this can be reverted since we already have a fix to signing in oie.

server/build.xml
```diff
@@ -1246,66 +1185,131 @@
...
+		<delete dir="${junit-html}" />
+
 		<mkdir dir="${junit-reports}" />
 		<mkdir dir="${code-coverage-reports}" />
-		
+		<mkdir dir="${junit-html}" />
+
 		<jacoco:coverage destfile="${code-coverage-reports}/jacoco.exec" xmlns:jacoco="antlib:org.jacoco.ant" exclclassloader="sun.reflect.DelegatingClassLoader:javassist.Loader" >
-			<junit haltonfailure="false" fork="true" forkmode="once">
+			<junit haltonfailure="false" fork="true" forkmode="perTest">
```
mgaffigan: Is this causing perf issues?  Tests felt a lot longer (11 minutes vs 3 minutes).

server/test/com/mirth/connect/model/PublicServerSettingsTest.java
mgaffigan, line 1: This seems like another of the `mirth_source` -> `message_source` changes.

mgaffigan (changes requested):
Compile failures:
- [x] [javadoc] error: invalid flag: -add-modules

New test-compile warnings:
- [x] [javac] warning: [options] --add-opens has no effect at compile time
- [x] [junit] OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
- [x] [junit] WARNING: package sun.misc not in java.base

Test failures:
- [x] [junit] TEST com.mirth.connect.donkey.server.channel.ChannelTest FAILED
- [x] [junit] TEST com.mirth.connect.donkey.server.data.jdbc.JdbcDaoTest FAILED
- [x] [junit] Test com.mirth.connect.client.core.ConnectServiceUtilTest FAILED
- [x] [junit] Test com.mirth.connect.model.PublicServerSettingsTest FAILED
- [ ] ~~[junit] Test com.mirth.connect.util.JavaScriptSharedUtilTest FAILED~~ (#184 opened - unrelated to this PR)

General concerns:
- [x] gitignore needs updated or build needs updated to avoid un-ignored untracked files after build (307 generated)
- [x] ~~If we're going to spend time testing a release, we should verify the versions of deps we're going to are best~~ (Getting this merged doesn't stop us from future updates).
- [x] A lot of these changes are formatting or whitespace.  Not strictly a problem, but the PR can be substantially minimized.

server/build.xml
mgaffigan: Perf issues were due to signing changes.  Reverted.  Change in perf is negligible.

server/build.xml
mgaffigan: Reverted

server/build.xml
mgaffigan: Reverted.

mgaffigan (approved):
Let's go!  Seems to be a step in the right direction.  Moves the build process to Java 17.  Seems to open and run as usual in Ballista.

client/ant-build.xml
```diff
@@ -347,6 +359,14 @@
...
 				<jvmarg value="-Xms128m" />
 				<jvmarg value="-Xmx2048m" />
+				<!-- https://stackoverflow.com/questions/54205486 -->
+				<jvmarg value="-Xshare:off" />
+				<jvmarg value="--add-opens=java.base/jdk.internal.misc=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.internal=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.providers=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/javax.sql.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql/java.sql=ALL-UNNAMED"/>
```
tonygermano: Are all of these changes necessary for the client?

command/build.xml
```diff
@@ -86,6 +98,14 @@
...
 				<jvmarg value="-Xms128m" />
 				<jvmarg value="-Xmx2048m" />
+				<!-- https://stackoverflow.com/questions/54205486 -->
+				<jvmarg value="-Xshare:off" />
+				<jvmarg value="--add-opens=java.base/jdk.internal.misc=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.internal=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.providers=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/javax.sql.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql/java.sql=ALL-UNNAMED"/>
```
tonygermano: Are these changes necessary?

donkey/build.xml
```diff
@@ -85,6 +97,14 @@
...
 				<jvmarg value="-Xms128m" />
 				<jvmarg value="-Xmx2048m" />
+				<!-- https://stackoverflow.com/questions/54205486 -->
+				<jvmarg value="-Xshare:off" />
+				<jvmarg value="--add-opens=java.base/jdk.internal.misc=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.internal=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/com.sun.rowset.providers=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql.rowset/javax.sql.rowset=ALL-UNNAMED"/>
+				<jvmarg value="--add-opens=java.sql/java.sql=ALL-UNNAMED"/>
```
tonygermano: It looks like these options were copy/pasted to all of the build files, and I don't think that is correct.

server/build.xml
```diff
@@ -228,9 +228,6 @@
 				<include name="*.xml" />
 			</fileset>
 		</copy>
-		<copy todir="${connectors.http}/lib" failonerror="false">
-			<fileset dir="${lib.extensions}/http" />
-		</copy>
 		<jar destfile="${connectors.http}/http-shared.jar" basedir="${classes}">
```
tonygermano, line 234: There are a lot of similar removals here. Technically the directories appear to not exist (I haven't checked all of them,) so there is likely no harm in removing them, but I'm sure the intent is for all of the plugin sections to be the same, whether they have additional libraries or not. 

Are these changes we want to keep?

server/build.xml
```diff
@@ -1191,24 +1089,24 @@
...
 		</javadoc>
 	</target>
 
 	<target name="create-derby-db" depends="compile">
 		<!-- re-create the embedded database -->
-		<delete dir="mirthdb" />
+		<delete dir="appdata/mirthdb" />
 		<java classname="com.mirth.connect.server.tools.ScriptRunner" fork="true" dir="${basedir}" failonerror="true">
+			<!-- Set Derby home so it creates mirthdb inside appdata -->
+			<jvmarg value="-Dderby.system.home=appdata" />
```
tonygermano: I'm not sure why these changes are here.

server/test/com/mirth/connect/connectors/dimse/DICOMDispatcherTest.java
```diff
@@ -100,7 +100,7 @@ public void testResponseData() throws DonkeyElementException {
         Response response = dispatcher.send(props, message);
         String responseData = response.getMessage();
 
-        String expectedResponseString = "<dicom>\n" + "<tag00000900 len=\"2\" tag=\"00000900\" vr=\"IS\">0</tag00000900>\n" + "</dicom>\n" + "";
+        String expectedResponseString = "<dicom><tag00000900 len=\"2\" tag=\"00000900\" vr=\"IS\">0</tag00000900></dicom>";
```
tonygermano, line 103: Do we know which change caused the newline character removal from the expected response?

server/test/com/mirth/connect/model/PublicServerSettingsTest.java
```diff
@@ -51,13 +51,13 @@ protected void configure() {
...
     public void defaultMetaDataColumnsTest() {
-        PublicServerSettings publicServerSettings = new PublicServerSettings(serverSettings);      
+        PublicServerSettings publicServerSettings = new PublicServerSettings(serverSettings);
         assertEquals(COLUMNS, publicServerSettings.getDefaultMetaDataColumns());
     }
     
     @Test
     public void queueBufferSizeTest() {
-        PublicServerSettings publicServerSettings = new PublicServerSettings(serverSettings);
+        com.mirth.connect.model.PublicServerSettings publicServerSettings = new com.mirth.connect.model.PublicServerSettings(serverSettings);
```
tonygermano: huh? why full package names?

server/test/com/mirth/connect/server/controllers/TestUtils.java
```diff
@@ -552,6 +552,19 @@ public static void fixMessageIdSequence(String channelId) throws Exception {
...
+                result.next();
+                int tableExists = result.getInt(1);
+                close(result);
+
+                // Create the table if it doesn't exist
+                if (tableExists == 0) {
+                    statement.execute("CREATE TABLE D_MESSAGE_SEQUENCES (LOCAL_CHANNEL_ID BIGINT NOT NULL PRIMARY KEY)");
+                }
+
                 result = statement.executeQuery("SELECT MAX(id) FROM d_m" + localChannelId);
```
tonygermano, line 555: connection is not database dependent, but these new queries are.

server/test/com/mirth/connect/server/servlets/WebStartServletTest.java
```diff
@@ -54,7 +54,10 @@ public void testDoGetCore() throws Exception {
 
 		webStartServlet.doGet(request, response);
 
-		assertEquals(CORE_JNLP.trim(), response.getResponseString().trim());
+		assertEquals(
+				CORE_JNLP.trim().replaceAll("\\s+", ""),
+				response.getResponseString().trim().replaceAll("\\s+", "")
+		);
 		assertEquals("application/x-java-jnlp-file", response.getContentType());
```
tonygermano, line 58: These repeated changes throughout the file to remove all whitespace do not seem appropriate.

tonygermano (changes requested):
17 commits and a lot of unrelated changes makes this somewhat messy and difficult to review (I am aware that much of it was probably externally generated.) It's difficult to tell which changes are necessary for the dependency updates and which are not. There are a lot of which I marked in my review as potentially being unnecessary changes. There are many more that are additional comments and whitespace changes which I did not mark, but probably don't belong in this PR (not that I have anything against comments, but they just bloat this already large PR.)

I think while we are changing the minimum required java version, we should do it correctly and close out https://github.com/OpenIntegrationEngine/engine/issues/18. I think this only entails modifying the ant javac tasks to include a `release="17"` attribute. It may be wise to store the release version in a build property as it will likely get updated from 17 to 21 in the not distant future now that version 25 is out.

See https://ant.apache.org/manual/Tasks/javac.html and https://stackoverflow.com/a/43103038

server/build.xml
mgaffigan: I agree they are unrelated to java 17 - but they add a bunch of warnings to the build output, making it hard to identify actual issues caused by Java 17.

I think these are beneficial changes, since there is not seemingly an ant option to avoid the warning otherwise.

server/test/com/mirth/connect/connectors/dimse/DICOMDispatcherTest.java
mgaffigan: I do not.

server/build.xml
mgaffigan: Neither am I.  Reverted.

server/test/com/mirth/connect/server/servlets/WebStartServletTest.java
mgaffigan: I agree that they are not ideal, but I think this was probably the correct engineering choice.  This is just a unit test and some copypasta hurts no-one in this context.

The hour I've spent looking into the details of this change was not beneficial.  See [attached](https://github.com/user-attachments/files/22608750/foo.xml) for the actual value received in the test.  The server seems to be adding inter-element whitespace for no reason I can deduce.  Also maybe switching EOL based on what OS it runs on.

I don't think there's a bug here.  I've moved the normalization to a function.

server/test/com/mirth/connect/server/controllers/TestUtils.java
mgaffigan: Agree.  Reverted.  I'm also not clear why this was required in the first place.

server/test/com/mirth/connect/model/PublicServerSettingsTest.java
mgaffigan: No idea.  Reverted.

donkey/build.xml
mgaffigan: Agree.  Reverted the add-opens.  `-Xshare:off` is required to avoid the warning mentioned in the comment.

command/build.xml
mgaffigan: Not seemingly.  Reverted the add-opens.  `-Xshare:off` is required to avoid the warning mentioned in the comment.

client/ant-build.xml
mgaffigan: Not seemingly.  Reverted the add-opens.  `-Xshare:off` is required to avoid the warning mentioned in the comment.

server/test/com/mirth/connect/connectors/dimse/DICOMDispatcherTest.java
mgaffigan: @tonygermano, do you view this as a blocker?

server/build.xml
mgaffigan: @tonygermano, do you view this as a blocker?

tonygermano: This is addressed in #200

server/test/com/mirth/connect/connectors/dimse/DICOMDispatcherTest.java
tonygermano: Not necessarily a blocker, but it would be good to know the source of this change in output to determine if it might affect other things which we need to warn people about.

tonygermano: This has to do with differences in xalan between java 8 and java 17, and not any of the changes in this PR.

tonygermano (approved):

PR #148, opened 2025-07-26
Fix warnings about deprecated constructors

mgaffigan:
Fixes javac build warnings present on Java 8 and Java 17.

### DIFF
--- a/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
+++ b/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
@@ -968,11 +968,11 @@ public class MessageBrowser extends javax.swing.JPanel {
                         }
 
                         if (!messages.hasNextPage()) {
-                            messages.setItemCount(new Long(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
+                            messages.setItemCount((long)(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
                             enableCountButton = false;
                         }
                     } else {
-                        messages.setItemCount(new Long((retrievedPageNumber - 1) * messages.getPageSize()));
+                        messages.setItemCount((long)((retrievedPageNumber - 1) * messages.getPageSize()));
                         enableCountButton = false;
                         pageNumberField.setText("0");
                     }
--- a/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
+++ b/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
@@ -968,11 +968,11 @@ public class MessageBrowser extends javax.swing.JPanel {
                         }
 
                         if (!messages.hasNextPage()) {
-                            messages.setItemCount((long)(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
+                            messages.setItemCount(Long.valueOf(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
                             enableCountButton = false;
                         }
                     } else {
-                        messages.setItemCount((long)((retrievedPageNumber - 1) * messages.getPageSize()));
+                        messages.setItemCount(Long.valueOf((retrievedPageNumber - 1) * messages.getPageSize()));
                         enableCountButton = false;
                         pageNumberField.setText("0");
                     }
--- a/donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
+++ b/donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
@@ -77,7 +77,7 @@ public class BufferedDao implements DonkeyDao {
 
     @Override
     public void commit(boolean durable) {
-        commit((Boolean) durable);
+        commit(Boolean.valueOf(durable));
     }
 
     private void commit(Boolean durable) {
--- a/donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
+++ b/donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
@@ -77,7 +77,7 @@ public class BufferedDao implements DonkeyDao {
 
     @Override
     public void commit(boolean durable) {
-        commit(new Boolean(durable));
+        commit((Boolean) durable);
     }
 
     private void commit(Boolean durable) {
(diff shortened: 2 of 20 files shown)


donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
```diff
@@ -77,7 +77,7 @@ public void commit() {
 
     @Override
     public void commit(boolean durable) {
-        commit(new Boolean(durable));
+        commit((Boolean) durable);
```
tonygermano: question: Is there a reason you chose casting+autoboxing vs `Boolean.valueOf(durable)`?

server/src/com/mirth/connect/connectors/jdbc/Column.java
```diff
@@ -91,7 +91,7 @@ public int hashCode() {
         int hashCode = 1;
         hashCode = 31 * hashCode + (name == null ? 0 : name.hashCode());
         hashCode = 31 * hashCode + (type == null ? 0 : type.hashCode());
-        hashCode = 31 * hashCode + (new Integer(precision)).hashCode();
+        hashCode = 31 * hashCode + precision;
```
tonygermano, line 94: This should be fine. I was unaware the hashcode of an Integer is the unwrapped int.

server/src/org/mozilla/javascript/NativeBoolean.java
```diff
@@ -101,7 +101,7 @@ public Object execIdCall(IdFunctionObject f, Context cx, Scriptable scope,
             return value ? "true" : "false";
 
           case Id_toSource:
-            return value ? "(new Boolean(true))" : "(new Boolean(false))";
+            return value ? "new Boolean(true)" : "new Boolean(false)";
```
tonygermano: This is part of Rhino, and probably shouldn't change.

server/src/org/mozilla/javascript/tools/debugger/SwingGui.java
```diff
@@ -787,7 +787,7 @@ static void setResizeWeight(JSplitPane pane, double weight) {
         try {
             Method m = JSplitPane.class.getMethod("setResizeWeight",
                                                   new Class[]{double.class});
-            m.invoke(pane, new Object[]{new Double(weight)});
+            m.invoke(pane, new Object[]{(double)weight});
```
tonygermano: cast is redundant. weight is already a double

server/test/com/mirth/connect/connectors/tcp/TcpDispatcherTest.java
tonygermano, line 1: My preference would be `Integer.valueOf(2)` over `(Integer)2` for clarity. This form appears several times throughout this file.

server/src/com/mirth/connect/server/controllers/DonkeyEngineController.java
```diff
@@ -966,11 +966,11 @@ public int compare(DashboardStatus o1, DashboardStatus o2) {
     }
 
     protected Long getSourceQueueSize(Channel channel) {
-        return new Long(channel.getSourceQueue().size());
+        return (long)(channel.getSourceQueue().size());
     }
 
     protected Long getDestinationQueueSize(DestinationConnector destinationConnector) {
-        return new Long(destinationConnector.getQueue().size());
+        return (long)(destinationConnector.getQueue().size());
```
tonygermano: I would prefer to see `Long.valueOf(channel.getSourceQueue().size())` vs `(long)(channel.getSourceQueue().size())`.

server/src/com/mirth/connect/util/MirthXmlUtil.java
```diff
@@ -125,7 +125,7 @@ public static String decode(String entity) {
                 start++;
                 radix = 16;
             }
-            Character c = new Character((char) Integer.parseInt(entity.substring(start), radix));
+            Character c = (char) Integer.parseInt(entity.substring(start), radix);
```
tonygermano, line 128: I believe this cast is ok since it is a narrowing conversion, which is then autoboxed.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
```diff
@@ -968,11 +968,11 @@ public void done() {
                         }
 
                         if (!messages.hasNextPage()) {
-                            messages.setItemCount(new Long(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
+                            messages.setItemCount((long)(((retrievedPageNumber - 1) * messages.getPageSize()) + messages.size()));
                             enableCountButton = false;
                         }
                     } else {
-                        messages.setItemCount(new Long((retrievedPageNumber - 1) * messages.getPageSize()));
+                        messages.setItemCount((long)((retrievedPageNumber - 1) * messages.getPageSize()));
```
tonygermano: should be able to use `Long.valueOf(long)` instead of casting to `(long)` and autoboxing.

tonygermano (changes requested):
I made several comments related to the auto-boxing commit. In general, where auto-boxing of primitives without casting is not enough, I would prefer to see the PrimitiveWrapper.valueOf() methods used over a cast. There are a few places where you have done it that way, and others where you have not.

I don't think we should make the security manager changes at this time. The security manager still works through java 23. According to https://openjdk.org/jeps/486, from java 24 on, it behaves as if a security manager is not installed, but the api methods have not yet been removed. Since we don't know who is currently using a security manager, I think we should not remove support for it without adequate warning. Obviously, if we eventually drop support for java versions < 24 there is no reason to keep it.

server/src/org/mozilla/javascript/NativeBoolean.java
mgaffigan: Yep.  Missed the parens when reverting the extra change.

donkey/src/main/java/com/mirth/connect/donkey/server/data/buffered/BufferedDao.java
mgaffigan: No.  `valueOf` makes more sense.  Updated.

tonygermano (approved):
Thanks. Changes all look good to me.

jonbartels (commented):
Is there a specific testing process we can use to observe these build warnings?

Just build it and read the output or is there a better way to filter through all the noise that `ant` spews out?

kayyagari (approved):

kpalang (approved):

PR #152, opened 2025-07-29
Fix warnings about java security manager

mgaffigan:
Fixes javac build warnings present on Java 8 and Java 17.

Split from #148

### DIFF
--- a/client/src/com/mirth/connect/client/ui/OSXAdapter.java
+++ b/client/src/com/mirth/connect/client/ui/OSXAdapter.java
@@ -52,11 +52,9 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
-import java.security.AccessController;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


tonygermano:
> @tonygermano, I think this is still safe to merge. JWS was removed in Java 11, and is non-default in Java 8. That's 10-years+ of this being a non-default issue.

Java WebStart utilized the SecurityManager, but SecurityManager is not a JWS specific technology. Some of these changes modify server classes as well as client classes.

I'm not saying that we shouldn't do this, only that we need to do it carefully. The technology is deprecated, but not dead, and someone could be using it (potentially with 2nd- or 3rd-party libraries.) There is no positive impact to this change other than removing a few compiler warnings, and there are potential negative impacts if we put it in the next release without giving adequate warning that it will be removed.

I appreciate you splitting it off from #148, which Kaur merged earlier today.

NicoPiel:
> someone could be using it (potentially with 2nd- or 3rd-party libraries.)

Yes, but it's not feasible for OIE to consider every possible deprecated use-case of disabled-by-default technology. Getting rid of stuff like this will only facilitate getting of Mirth's rat tail of ancient dependencies and technologies in the future.

PR #96, opened 2025-05-21
Remove simplesender

rogin:
For Issue #88

Searched the project for "simplesender". Removed its root folder then references in .gitignore.

Verified a successful build using Oracle JDK 8.

PR #163, opened 2025-08-14
Issue #88 remove SimpleSender

### LINKED ISSUE #88: Remove simplesender
The project located at `simplesender` is a wholly separate project.

It has not been updated since 2012 and has only one author. 

The tool appears to support:
* sending HL7 messages from strings or files
* making DB queries with JDBC

The tool has only two references in the old forums:
* https://forums.mirthproject.io/forum/mirth-connect/support/11871-mirth-2-x-projects
* https://forums.mirthproject.io/forum/mirth-connect/support/1067-llplistener-timeout-when-bufferedreader-readline

simplesender should be removed entirely

jonbartels:
Resolves #88 

The `simplesender` module is never used. There were also references to it in `.gitignore` that I removed.


pacmano1 (approved):

tonygermano (changes requested):
@jonbartels I meant to assist with testing against a running mirth server. Remember there are also a lot of tests we don't run automatically because they require a running mirth server, and we don't know exactly how those tests were used internally at NG either.

Regardless, this PR is a duplicate of https://github.com/OpenIntegrationEngine/engine/pull/96

Issue #156, opened 2025-08-06
[BUG] NextGen SSL Advertisement Hidden in HTTP Connector and SOAP Connector


jonbartels:
**Describe the bug**
https://github.com/OpenIntegrationEngine/engine/blob/main/client/src/com/mirth/connect/connectors/ws/SSLWarningPanel.java 

https://github.com/OpenIntegrationEngine/engine/blob/main/client/src/com/mirth/connect/connectors/http/SSLWarningPanel.java

This code shows an advertisement when the user enters "https" in the address for the HTTP or SOAP connectors.

This is lame and needs to be either removed or point to OIE documentation on using SSL. The scope here is only to remove or alter the warning. The scope should not include any changes to the HTTPS behaviors of OIE

<img width="654" height="508" alt="Lame Advertisement" src="https://github.com/user-attachments/assets/00f54385-71ab-42f9-a0b2-5ca0a14dee60" />

PR #164, opened 2025-08-15
Issue #156 - Change SSL help text to a Constant, remove ads

jonbartels:
Issue #156 - Change SSL help text to a Constant, remove NextGen branding and ads from the message


kayyagari:
> I understand the desire to not want to duplicate code, but I don't know if UIConstants is the right place to put connector specific data.

IMO, It is all on the client UI and never gets into anything connector specific on the server, so this is acceptable.

pacmano1 (approved):

jonbartels:
I put it in UI Constants because multiple places showed this help text. 

There isn't any other common class to both WS and HTTP connectors that I could find.

It is help text shown in the UI, so UIConstants seemed like the best choice.

This isn't just duplicated code, its a whole block of HTML. Duplicating small strings maybe but an HTML block should never have been copy-pasted like the original.

kryskool (approved):

mgaffigan (approved):

tonygermano (approved):

pacmano1:
> I’m late to the party here, but I think we should completely remove this warning and the scary icon/s that goes with it. There is no warning for http traffic (nor am I advocating for one here) which is arguably higher risk.
> 
> When using https, we have reasonable defaults and the connection is reasonably secure. Further, the prior or the proposed warning does not give the user any beneficial or actionable information. There is no reason to believe that anything but the default cert store would be used and there is nothing inherently harmful by doing so.
> 
> Also, mTLS _is_ supported, it just needs to be configured with the client cert/key. We don’t currently provide a way to do that, but it can be done at the JVM level and the connection will pick it up and use it just fine.
> 
> If we feel we need a message, then perhaps a link to the documentation for advanced configurations.

Perhaps.   I think we need to be sensitive to conditioned behavior for many users using this engine for many years.

I would not remove it for this reason - it is perhaps a gentle reminder.

client/src/com/mirth/connect/client/ui/UIConstants.java
```diff
@@ -129,4 +129,13 @@ public class UIConstants {
...
     public static final String USER_PREF_KEY_BACKGROUND_COLOR = "backgroundColor";
+
+    // SSL Warning Text
+    // This text is used in the SSL Warning Panel for HTTP and WS connectors.
+    public static final String SSL_WARNING_TEXT = "<html>\n" +
+            "<body class=\"ssl-warning-panel\"><b><span style=\"color:#404040\">Important Notice:</span></b> The default system\n" +
+            "certificate store will be used for this connection. As a result, certain security options are not available and mutual\n" +
+            "authentication (two-way authentication) is not supported.\n" +
+            "</body>\n" +
+            "</html>";
```
tonygermano: Why have there been `\n` characters inserted into the string?

mgaffigan: I can't answer for @jonbartels, obviously, but I often put them in in that way to avoid the source code appearing different from the value of the constant.  They are interpreted in HTML as whitespace - which avoids "The default systemcertificate..." vs "The default system certificate...".

Whether that's the reason or not, it seems like this PR moves us to a better spot and doesn't appear to introduce any new problems.

jonbartels: The "\n" makes the text render in multiline so its easier to read.

It's a habit and what is generated by an autoformatter. Like Mitch said it doesnt affect the output

pacmano1 (changes requested):
```
This connection uses the java system trust store, which limits available TLS security options. \n
Client certificate selection is not available in the UI when using the system store, preventing mutual TLS (mTLS) configuration through this interface.
```

Only that for the text.

client/src/com/mirth/connect/connectors/ws/WebServiceSender.java
```diff
@@ -80,10 +80,10 @@
 import com.mirth.connect.model.Connector.Mode;
 import com.mirth.connect.model.converters.ObjectXMLSerializer;
 import com.mirth.connect.util.ConnectionTestResponse;
+import com.mirth.connect.client.ui.UIConstants;
 
 public class WebServiceSender extends ConnectorSettingsPanel {
 
-    protected static final ImageIcon ICON_LOCK_X = new ImageIcon(Frame.class.getResource("images/lock_x.png"));
```
tonygermano, line 86: I think this same change needs to be made in the HttpSender

tonygermano (approved):
I pushed [changes](https://github.com/OpenIntegrationEngine/engine/pull/164/commits/3da3d3c24922596673817cca142337832aea8c8a) to this PR. The previous commit did not work because of the issue described in the commit message. The revised version looks like the screenshot below. It incorporates some of @pacmano1's suggestions for the text of the message, and the icon change from @jonbartels's  previous commits. The new tooltip text will appear when hovering over either of the sections marked in red.

Image

This decoration can be overridden/removed by plugins with proper TLS support.

pacmano1 (approved):

tonygermano (approved):

Issue #177, opened 2025-08-27
[IDEA] Pimp my OIE Administrator


kryskool:
**Is your feature request related to a problem? Please describe.**
On a new OIE installation, we have the same color as the mirth connect installation

**Describe your use case**
When migrating from Mirth Connect to OIE, the GUIs should have a different color to differentiate them at first glance.

**Describe the solution you'd like**
Change the default color for the new installation.

**Describe alternatives you've considered**
no opinion on what color to use

**Additional context**

<img width="695" height="654" alt="Image" src="https://github.com/user-attachments/assets/c4059fce-f3f4-4e00-ba24-81011df0613f" />

PR #178, opened 2025-09-05
Update ServerSettings.java

mhokanson:
Update the default color for the Administrator window to further differentiate OIE from Mirth Connect.

[Issue 177](https://github.com/OpenIntegrationEngine/engine/issues/177)


mgaffigan (approved):

mhokanson:
Here's a side-by-side comparison of the change.

Again, this only affects the out-of-the-box default color and won't affect custom colors already set and saved (example: Production in "Warning Red").

image

kryskool (approved):

tonygermano (commented):
I think we should give this one another week or so to sit to see if it gets any additional feedback from the community.

PR #192, opened 2025-10-18
Add standard Jackson JSON media type to admin API

### LINKED ISSUE #188: [BUG] OpenAPI definition does not match response
**Describe the bug**
The OpenAPI schemas do not match the responses from the API's themselves.  They seem to differ in at least two ways:

1. Missing root element<br/>Each API response includes the serialized type name which is omitted from the declared schema
2. Incorrect date-time formatting<br/>API responses declare date-time fields, which are defined in OpenAPI3 to be compliant with RFC3339 §5.6 (ISO8601 style strings).  API Endpoints are actually returning an object with a ticks field.

**To Reproduce**
Call /api/openapi.json to get schema
Call /api/users/current to get actual.

**Expected behavior**
Schema and actual should match

**Actual behavior**
Schema and actual do not match

**Screenshots**

Image

**Environment**
 - Windows 11
 - Zulu 17
 - 6865d88e49f998fad978381400b2a3f0bfb0e496

mgaffigan:
Adds application/mirthapi+json as a supported content-type for the admin API's that serialize or deserialize objects.  

- Avoids the round-trip through XML as currently performed by ObjectJSONSerializer/JsonMessageBodyReader
- Matches the OpenAPI spec, allowing auto-generated clients in most languages
- Probably more performant, but not really the point.

`application/json` keeps the historical behavior - this functionality is opt-in.

Closes #188

### DIFF
--- a/server/src/com/mirth/connect/client/core/api/servlets/AlertServletInterface.java
+++ b/server/src/com/mirth/connect/client/core/api/servlets/AlertServletInterface.java
@@ -37,6 +37,7 @@ import com.mirth.connect.client.core.Operation.ExecuteType;
 import com.mirth.connect.client.core.Permissions;
 import com.mirth.connect.client.core.api.BaseServletInterface;
 import com.mirth.connect.client.core.api.MirthOperation;
+import com.mirth.connect.client.core.api.ApiContentTypes;
 import com.mirth.connect.client.core.api.Param;
 import com.mirth.connect.model.ChannelHeader;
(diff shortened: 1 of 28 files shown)


ChristopherSchultz:
If you are going to standardize, why use `application/mirthapi+json` which is yucky. Just use `application/json`. There shouldn't be anything Mirth-specific in the syntax of these data; the custom subtype is just vanity and should be removed.

PR #194, opened 2025-10-21
Add more module options to default vmoptions file

kpalang:
Add mode `--add-opens` directives to the default vmoptions.


jonbartels (approved):
LGTM

mgaffigan (approved):
I'd like to be removing opens - not adding them, but I don't see a downside this PR.

tonygermano (changes requested):
I'm not sure if this is the correct thing to do or not. As it is, I think this file only contains the minimum required by the java code in this project.

Do we want to be adding modules just because they could potentially be used in user code or should those just be added to the user's custom options as required?

mgaffigan:
@kpalang, can you document what is benefitting from these additional opens?

mgaffigan (approved):
LGTM

PR #201, opened 2025-11-07
Update CONTRIBUTING.md to enhance build instructions and component-sp…

charEtoile:
Update CONTRIBUTING.md to enhance build instructions


CONTRIBUTING.md
```diff
@@ -29,14 +29,109 @@ git checkout -b feature/your-feature-name
...
+- Build Client
+- Build Manager
+- Build CLI (Command Line Interface)
+- Run all tests
+- Create the complete setup in `server/setup/`
+
+**Full Build (Development)**
+```bash
+cd server
+ant -f mirth-build.xml -DdisableSigning=true
```
tonygermano: We have also added a `-DdisableTests=true` property that can be set to skip running tests as part of the build if you would like to add a section about that. See https://github.com/OpenIntegrationEngine/engine/pull/141

CONTRIBUTING.md
```diff
@@ -29,14 +29,109 @@ git checkout -b feature/your-feature-name
...
+cd server
+ant -f mirth-build.xml
+```
+
+**Create Distribution Package**
+```bash
+cd server
+ant -f mirth-build.xml dist
+```
+
```
tonygermano, line 113: Maybe add a section about starting the server after building? I see further down you indicated that `server/setup` is where the application is "installed" post-build. As of https://github.com/OpenIntegrationEngine/engine/pull/119 there will be an `oieserver` bash script for linux/mac developers and `oieserver.ps1` powershell script for windows developers placed in `server/setup` to launch the server with that will correctly pick up the vmoptions.

CONTRIBUTING.md
charEtoile: Hi, I have added a section for these tests.

CONTRIBUTING.md
charEtoile: Hi, actually, I tried to start the client ... and the old java webstart was an issue. So i decided to upgrade the whole thing to a more recent JAVA, the 21. I updated the `CONTRIBUTING.md` to give running instructions. I added an `oieclient` script based on the `oeiserver` script.
[https://github.com/charEtoile/engine/blob/update_sdk_to_21/CONTRIBUTING.md](https://github.com/charEtoile/engine/blob/update_sdk_to_21/CONTRIBUTING.md)
I know that current tests fail, i am investigating.

tonygermano: As things currently stand, the project must compile with java 8, but the oieserver script requires the server is started with java 17 or higher. There are a couple work in progress PRs that will require java 17 for the build version, and 17 or higher for the runtime version.

charEtoile: OK.
I reverted back to Java 8 and added instructions for Java 17 runtime. All build tests pass.
I also added the oieclient script. Hope this is enough.

server/basedir-includes/oieserver
```diff
@@ -277,5 +277,4 @@ JAVA_OPTS=("${VMOPTIONS[@]}"
 
 # Launch Open Integration Engine (as this PID with exec)
 echo "Starting Open Integration Engine..."
-echo "$FINAL_JAVA_CMD ${JAVA_OPTS[*]}"
```
tonygermano, line 280: I don't think this line should be removed. Potentially it could require a verbose flag, but I think that change would belong in a different PR.

server/conf/custom.vmoptions
```diff
@@ -22,3 +22,7 @@
 #
 # Any lines which do not match the special directives above will be passed through directly as
 # arguments to the java binary.
+
+# Use the Java version specified in .sdkmanrc (located in project root)
+# This ensures server runtime matches the build environment
+-java-cmd ${HOME}/.sdkman/candidates/java/17.0.17.fx-zulu/bin/java
```
tonygermano, line 28: This should not be added here. This file is included in the full release where the user is likely to not be using sdkman and does not have access to the project root.

tonygermano (changes requested):
I haven't given a full review to the oieclient script yet. I'm not confident it is included in the correct place, but I can see how it could be useful during development, as there are other similar scripts already floating around.

I think rather than `/server/base-includes` that `/tools/client-launcher` or something similar may be a more appropriate location, but I'll wait for others to weigh in on that. I don't think it's a script that we want to ship with the installer, as it's very uncommon to launch the client from the server host in production, especially for a non-windows server. Typically, once the server is launched, a standard client launcher tool is used to connect. See https://docs.openintegrationengine.org/launchers/. Most of us use Ballista during development.

I also like the additional detail you've added about running the client since my last review, but I wonder if that is too much for the contributing guide and would be better placed on the documentation site linked above. I requested the information on starting the server since that is slightly different when running the build yourself vs installing the release, but the launching the client does not need to be different than in production. I'll let others provide feedback on this as well.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
 ```
 
 ### 5. Install Tooling
-OIE specifies the working versions of Java and Ant in [.sdkmanrc](./.sdkmanrc). To take advantage of this, install [SDKMAN](https://sdkman.io/) and run `sdk env install`
-in the project's root directory.
```
mgaffigan, line 30: Why are we moving away from `sdk env install`?

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+- **Runtime**: Java 17 or higher (required for server and client execution)
+
+#### Install Java Versions with SDKMAN
+
+Install [SDKMAN](https://sdkman.io/) and use it to manage both Java versions:
+
+```bash
+# Install SDKMAN (if not already installed)
+curl -s "https://get.sdkman.io" | bash
+source "$HOME/.sdkman/candidates/java/current/bin/java"
```
mgaffigan, line 44: I would not repeat installation guidance given that it may change - perhaps a link to the relevant documentation.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+# Install Java 17 with JavaFX for runtime (required for client GUI)
+sdk install java 17.0.17.fx-zulu
+
+# Install Apache Ant
+sdk install ant 1.10.14
+```
+
+**Set build-time Java:**
+```bash
+sdk use java 8.0.442.fx-zulu
```
mgaffigan, line 58: This should not be required with `sdk env install`

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+```
+
+**Verify versions:**
+```bash
+java -version  # Should show Java 8
+ant -version   # Should show Ant 1.10.14
+```
+
+> **Important:** After building with Java 8, you'll need Java 17+ with JavaFX (`17.0.17.fx-zulu` or higher) to run the server and client. The runtime Java is specified in `conf/custom.vmoptions` for the server and `oieclient.vmoptions` for the client.
```
mgaffigan, line 67: Server runtime java can also be configured by `OIE_JAVA_HOME` or by other means - not sure this should be in the getting started.  Probably should be in the documentation for oieserver or similar.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+```bash
+cd server
+ant -f mirth-build.xml
+```
+
+**Create Distribution Package**
+```bash
+cd server
+ant -f mirth-build.xml dist
+```
```
mgaffigan, line 112: This is all good information, but probably does not need to be in the getting started.  Perhaps in build system documentation.  Not sure about including normative build speed - it's a figure that is unlikely to be updated over time, and will get out of date.

CONTRIBUTING.md
mgaffigan: @charEtoile, I like the idea, but please consider pulling oieclient.sh into a separate PR.  CONTRIBUTING.md certainly needs updates, but the resounding runtime pattern is to use a launcher to start the client.  Changing that is a much larger change than documentation.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+```bash
+cd server
+cp -r setup openintegrationengine
+tar czf ../openintegrationengine.tar.gz openintegrationengine
+rm -rf openintegrationengine
+```
+
+Alternatively, install GNU tar on macOS:
+```bash
+brew install gnu-tar
```
mgaffigan, line 210: +1 for `brew install gnu-tar`.  More options = more problems.  I don't think release procedures need to be in the quick start at all, let alone for Mac OS specifically.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+
+**Note:** Both scripts automatically discover Java in this priority order:
+1. `OIE_JAVA_PATH` environment variable (override)
+2. `-java-cmd` directive in respective `.vmoptions` file (preferred)
+3. `JAVA_HOME` environment variable
+4. `java` command in system PATH
+
+#### Packaging
+
+After building, you can create a distribution tarball:
```
mgaffigan, line 187: Note: this does not include installers.  Installers are published using install4j, which requires a license to build.

CONTRIBUTING.md
```diff
@@ -26,17 +26,232 @@ git checkout -b feature/your-feature-name
...
+rm -rf openintegrationengine
+```
+
+Alternatively, install GNU tar on macOS:
+```bash
+brew install gnu-tar
+gtar czf openintegrationengine.tar.gz -C server/ setup --transform 's|^setup|openintegrationengine/|'
+```
+
+#### Build Individual Components
```
mgaffigan, line 214: I don't think this is needed for quick-start.  Maybe in some IDE configuration documentation under.  Quick start should be focused on getting started quickly, not efficiency.  Deliver value to the contributor, and they will keep reading on how to become efficient.

CONTRIBUTING.md
charEtoile: I will split the PR for the oeiclient.s script and its associated documentation. I focus mainly on dev, not necessarily on production, so maybe I should state somewhere that the script is provided as a convenience for developers rather than for use in production.

server/basedir-includes/oieserver
charEtoile: I'll see if I can add a verbose flag.

PR #175, opened 2025-08-25
Login and client command line refactoring

mgaffigan:
- Refactor to allow alternate implementations of LoginPanel
- Moved handleLoginSuccess to Mirth client since it is actually setting up the process - not really doing anything with the panel
- Make client command line testable

In support for future merge of [mgaffigan:feature/add-oidc-auth](https://github.com/OpenIntegrationEngine/engine/compare/main...mgaffigan:oie:feature/add-oidc-auth)

PR #202, opened 2025-11-14
refactor client command line parsing and LoginStatus success check

tonygermano:
Extracted the first two commits from #175 to make it a more digestible chunk. No changes except to file headers.

The new files were missing headers, so I added them in SPDX format. I also updated the headers of the remaining files in the commits marking the copyright holders of each file as I felt appropriate. The changes in these commits I do not think warranted a copyright addition for file LoginPanel.java, however, I think previous changes to this file did, as noted in the commit message.

### DIFF
--- a/client/src/com/mirth/connect/client/ui/LoginPanel.java
+++ b/client/src/com/mirth/connect/client/ui/LoginPanel.java
@@ -1,11 +1,6 @@
-/*
- * Copyright (c) Mirth Corporation. All rights reserved.
- * 
- * http://www.mirthcorp.com
- * 
- * The software in this package is published under the terms of the MPL license a copy of which has
- * been included with this distribution in the LICENSE.txt file.
- */
+// SPDX-License-Identifier: MPL-2.0
+// SPDX-FileCopyrightText: Mirth Corporation
+// SPDX-FileCopyrightText: 2025 Mitch Gaffigan and Tony Germano
 
 package com.mirth.connect.client.ui;
 
@@ -451,7 +446,7 @@ public class LoginPanel extends javax.swing.JFrame {
                     }
 
                     // If SUCCESS or SUCCESS_GRACE_PERIOD
-                    if ((loginStatus != null) && ((loginStatus.getStatus() == LoginStatus.Status.SUCCESS) || (loginStatus.getStatus() == LoginStatus.Status.SUCCESS_GRACE_PERIOD))) {
+                    if (loginStatus != null && loginStatus.isSuccess()) {
                         if (!handleSuccess(loginStatus)) {
                             LoginPanel.getInstance().setVisible(false);
                             LoginPanel.getInstance().initialize(PlatformUI.SERVER_URL, PlatformUI.CLIENT_VERSION, "", "");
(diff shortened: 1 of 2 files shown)


client/src/com/mirth/connect/client/ui/CommandLineOptions.java
```diff
@@ -0,0 +1,103 @@
...
+        String protocols = "";
+        String cipherSuites = "";
+
+        if (args == null) {
+            args = new String[0];
+        }
+
+        if (args.length > 0) {
+            server = args[0];
+        }
```
kpalang, line 36: nitpick: new line between blocks

client/src/com/mirth/connect/client/ui/CommandLineOptions.java
```diff
@@ -0,0 +1,103 @@
...
+        if (args.length > 1) {
+            version = args[1];
+        }
```
kpalang, line 39: nitpick: new line between blocks

client/src/com/mirth/connect/client/ui/CommandLineOptions.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+                if (args.length > 6) {
+                    password = args[6];
+                }
+            } else {
+                // <server> <version> <username> [<password> [-ssl [<protocols> [<ciphersuites>]]]]
+                username = args[2];
+                if (args.length > 3) {
+                    password = args[3];
+                }
```
kpalang, line 60: nitpick: new line between blocks

client/src/com/mirth/connect/client/ui/CommandLineOptions.java
```diff
@@ -0,0 +1,103 @@
...
+        return password;
+    }
+
+    public String getProtocols() {
+        return protocols;
+    }
+
+    public String getCipherSuites() {
+        return cipherSuites;
+    }
```
kpalang, line 102: suggestion: I would love for this class to either use Lombok's classwide `@Getter` or better yet, be a `record`. Although `record`s came about in Java 14...

tonygermano: I had the same thought about records, but we could probably just hold this PR until after the java 17 upgrade.

mgaffigan: The desire to use a feature not yet available to the codebase for no difference in behavior does not seem to be a good reason to delay a PR.

Here's the checklist in my head:

*Does this PR deliver value consistent with the project goals?*
Yes, it simplifies some code and is a precursor to a valuable feature.

*Is the code understandable/maintainable?*
Yes, though records may be more succinct, it is unremarkable code that any dev should be able to understand. (I would not say the same of AOP, since I don't like magic)

*Does it adhere to the patterns present in the repo?*
Yes, getters are rife within the repo. (ignoring separate discussions about whitespace)

*Is the change likely to introduce issues?*
No, to the contrary, this PR adds unit test coverage.

Consider opening a refactoring issue to track updating all the legacy classes to records.

kpalang: Mitch is correct here

client/src/com/mirth/connect/client/ui/CommandLineOptions.java
tonygermano: I'm actually going to disagree on this. I don't think it would improve readability and would just add more scrolling. Is there a style guide that recommends separating consecutive if statements with blank lines? If we keep it this way, the blank line on line 33 should probably be removed for consistency.

Issue #207, opened 2025-11-19
[BUG] SimpleDateFormat is not thread-safe


NicoPiel:
`SimpleDateFormat` is not thread-safe. Ex: https://stackoverflow.com/questions/4021151/java-dateformat-is-not-threadsafe-what-does-this-leads-to

Suggestion: use `FastDateFormat` instead: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/FastDateFormat.html

PR #208, opened 2025-11-19
Replaced SimpleDateFormat with DateTimeFormatter wherever thread-critical

NicoPiel:
Solves #207

### DIFF
--- a/server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
+++ b/server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
@@ -9,23 +9,24 @@
 
 package com.mirth.connect.client.core.api.providers;
 
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Type;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.util.Calendar;
-
 import javax.inject.Singleton;
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.ext.ParamConverter;
 import javax.ws.rs.ext.ParamConverterProvider;
 import javax.ws.rs.ext.Provider;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Calendar;
 
 @Provider
 @Singleton
 public class CalendarParamConverterProvider implements ParamConverterProvider {
-
-    private static final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 
     @Override
     public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
@@ -38,10 +39,16 @@ public class CalendarParamConverterProvider implements ParamConverterProvider {
                     }
 
                     try {
-                        Calendar date = Calendar.getInstance();
-                        date.setTime(format.parse(value));
-                        return (T) date;
-                    } catch (ParseException e) {
+                        // Parse the incoming string as a ZonedDateTime using the given pattern
+                        ZonedDateTime zdt = ZonedDateTime.parse(value, FORMATTER);
+
+                        Calendar calendar = Calendar.getInstance(java.util.TimeZone.getTimeZone(zdt.getZone()));
+                        calendar.setTimeInMillis(zdt.toInstant().toEpochMilli());
+
+                        @SuppressWarnings("unchecked")
+                        T result = (T) calendar;
+                        return result;
+                    } catch (DateTimeParseException e) {
                         throw new ProcessingException(e);
                     }
                 }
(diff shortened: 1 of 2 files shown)


mgaffigan (changes requested):
I agree with changing:
- server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
- server/src/com/mirth/connect/plugins/serverlog/ServerLogItem.java

but the others are all allocated on the stack - and have no possibility for cross-thread use.  Changing DateUtil in particular seems unnecessarily risky.

Also, is there a reason we're jumping to Apache FastDateFormat instead of Java 8 DateTimeFormatter?

NicoPiel:
> I agree with Mitch
> 
> > Also, is there a reason we're jumping to Apache FastDateFormat instead of Java 8 DateTimeFormatter?
> 
> The core Java DateTimeFormatter is rock solid.

That's why it's implemented :)

server/src/com/mirth/connect/plugins/serverlog/ServerLogItem.java
```diff
@@ -10,14 +10,17 @@
 package com.mirth.connect.plugins.serverlog;
 
 import java.io.Serializable;
-import java.text.SimpleDateFormat;
+import org.apache.commons.lang3.time.FastDateFormat;
```
mgaffigan: Remove unused import.

server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
```diff
@@ -38,10 +41,18 @@ public T fromString(String value) {
...
+
+                        // Convert to a Calendar in the same zone
+                        Calendar calendar = Calendar.getInstance();
+                        calendar.setTime(Date.from(zdt.toInstant()));
+                        calendar.setTimeZone(java.util.TimeZone.getTimeZone(zdt.getZone()));
+
+                        @SuppressWarnings("unchecked")
+                        T result = (T) calendar;
+                        return result;
+                    } catch (DateTimeParseException e) {
```
mgaffigan, line 51: This is a fair bit of work being done to get back to `Calendar`.  I think it has been done correctly, but since we're still using `Calendar` in the interface - I do question whether it is worth it to round-trip through java.time vs moving the construction of `SimpleDateFormat` inside the converter.  Long time ideal would be to update the interface to avoid accepting `Calendar` - but I don't think that is worth doing right now.

I'm not blocking on this one: I think the conversion through java.time is being done correctly.

NicoPiel:
> Please omit the unused import, then I'll be good.

Done.

mgaffigan (approved):

jonbartels (dismissed):

server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
```diff
@@ -9,27 +9,30 @@
...
 public class CalendarParamConverterProvider implements ParamConverterProvider {
-
-    private static final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    private static final DateTimeFormatter FORMATTER =
+            DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 
     @Override
     public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
-        if (rawType.getName().equals(Calendar.class.getName())) {
+        if (rawType == Calendar.class) {
```
tonygermano: Are we sure this is a safe change?

mgaffigan: Good point.  We do have several ClassLoaders in server - not sure it is a problem, but I don't see value in the change vs comparing name.

NicoPiel: Two classes may have the same name if they are loaded by different `ClassLoader`s. If we want to simply compare types, this way is safer.
See https://cwe.mitre.org/data/definitions/486.html

tonygermano: What makes me question the change is that
1. It was written to compare names in the first place.
2. At least on the server side there are many different ClassLoaders. I'm less familiar with the situation on the client side.
3. This area of the code is linked to serialization, which is an indicator the class may not have been instantiated through normal means.

As Mitch said, it might be fine, but I don't know for sure without doing a lot more digging.

I don't believe the existing code is a safety issue, as the name of the class is only used to drive whether this converter has support for that object type, and the encapsulated Calendar class is never accessed or used in the return value. I think a major difference between the existing code and the link you shared is that we are being passed a `java.lang.Class` directly, and not calling `.getClass()` on a potentially untrusted object to determine the type of that object.

NicoPiel: Agreed, reverted.

mgaffigan (dismissed):

tonygermano (dismissed):

kayyagari (dismissed):

server/src/com/mirth/connect/client/core/api/providers/CalendarParamConverterProvider.java
tonygermano: suggestion: regarding the number of steps to get to a Calendar, I believe this will produce the same result, and it avoids creating an intermediate `java.util.Date`. `TimeZone` could also be imported.

```java
Calendar calendar = Calendar.getInstance(java.util.TimeZone.getTimeZone(zdt.getZone()));
calendar.setTimeInMillis(zdt.toInstant().toEpochMilli());
```

I'm not blocking on this either.

NicoPiel: No, you're completely right. Your approach also minimises potential race conditions. Implemented!

kayyagari (dismissed):

mgaffigan (dismissed):

jonbartels (approved):

PR #214, opened 2025-12-01
Adds Gradle build that delegates to existing Ant build

### LINKED ISSUE #52: [IDEA] Introduce Dependency Management (e.g. Gradle or Maven)

NicoPiel:
Fixes #52

Test with `./gradlew clean build -PdisableTests=true -PdisableSigning=true `

Adds an initial Gradle build system that delegates to the existing Ant build to enable gradual migration and tooling improvements.

Highlights:
- Introduces a root Gradle configuration that imports and delegates to the project's Ant build targets, avoiding immediate rewrite of existing build logic.
- Adds the Gradle wrapper and supporting files (wrapper properties, jar, and scripts) so the repository can be built consistently without requiring a local Gradle install.
- Adds per-subproject Gradle stub files to make the repository a multi-project Gradle layout and allow incremental conversion of subprojects.
- Adds Gradle metadata/configuration (gradle.properties, dependency platform TOML)
- Adds .gitattributes and .gitignore tweaks and normalizes an IDE classpath file to remove trailing encoding differences.
- Configures Ant-to-Gradle interoperability (ant taskdefs and properties) and maps Ant targets to avoid conflicts with Gradle built-in tasks.

Why:
- Allows us to start using Gradle tooling and the wrapper for builds, CI, and IDEs while preserving the working Ant build.
- Supports incremental migration of subprojects to Gradle without disrupting existing build behavior.

Notes / Next steps:
- Subprojects remain backed by the Ant build; plan incremental conversion of individual modules to native Gradle builds.
- Verify CI pipelines and developer SDK setups pick up the wrapper and SDKMAN gradle setting.


client/build.gradle.kts
mgaffigan, line 1: Should the subprojects be created in the first commit?  Seems like this would imply and allow them to be run in gradle - which will have no effect.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.1-all.zip
```
mgaffigan, line 3: Do we want `-all.zip`?  This is apparently intended for offline use.  Default initialize gradle script uses `-bin.zip`.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,8 @@
+# This file was generated by the Gradle 'init' task.
+# https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format
+
+[versions]
+guava = "33.4.5-jre"
+
+[libraries]
+guava = { module = "com.google.guava:guava", version.ref = "guava" }
```
mgaffigan, line 8: Guava doesn't seem to be referenced?

settings.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * The settings file is used to specify which projects to include in your build.
+ * For more detailed information on multi-project builds, please refer to https://docs.gradle.org/8.14/userguide/multi_project_builds.html in the Gradle documentation.
+ * This project uses @Incubating APIs which are subject to change.
+ */
+
+plugins {
+    // Apply the foojay-resolver plugin to allow automatic download of JDKs
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.10.0"
+}
```
mgaffigan, line 12: Not used?

settings.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+
+rootProject.name = "OIE"
```
mgaffigan: Naming convention for gradle seems to be all-lowercase separated by hyphens.  I suggest oie or open-integration-engine.

build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Pass Gradle properties to Ant
+project.findProperty("disableSigning")?.let {
+    ant.properties["disableSigning"] = it.toString()
+}
+
+project.findProperty("disableTests")?.let {
+    ant.properties["disableTests"] = it.toString()
+}
```
mgaffigan, line 64: Are there other properties we need to be passing?

mgaffigan (changes requested):
Other issues to address
- CI build scripts need updated (.github/workflow/...)
- Readme/Contributor documents need updated
- ant mirth-build build target needs to be updated to block build outside of gradle with migration guidance in error message

client/build.gradle.kts
NicoPiel: Since it has no effect, but lays the foundation for future migration work, I think it's fine, but not necessary.

gradle/wrapper/gradle-wrapper.properties
NicoPiel: This is what `gradle init` generated. I can use the bin version, though.

gradle/libs.versions.toml
NicoPiel: Also what `gradle init` generated.

settings.gradle.kts
NicoPiel: Also generated by `gradle init`

build.gradle.kts
NicoPiel: Good question. I'm not sure. These are the only two I use and the only two available in `mirth-build.xml`, afaia.

gradle/libs.versions.toml
NicoPiel: It may be something Gradle uses internally.

gradle/wrapper/gradle-wrapper.properties
mgaffigan: If it is from gradle init, then I'm fine with it.

.sdkmanrc
```diff
@@ -2,3 +2,4 @@
 # Add key=value pairs of SDKs to use below
 java=17.0.17.fx-zulu
 ant=1.10.14
+gradle=9.2.1
```
tonygermano: I don't think this change is necessary. The gradle version for the project is handled by the gradle wrapper.

settings.gradle.kts
tonygermano: agree with the lowercase. I think eventually the project name gets included in the jar filenames, so I'd say "oie"  for brevity.

build.gradle.kts
tonygermano: I know JUnit supports some properties for running specific tests instead of the whole suite, but I doubt anyone actually uses that. I have in the past, but I would need to look them up to even identify what they are called, because it's not something I do frequently. I wouldn't hold up this PR for that.

tonygermano: https://github.com/OpenIntegrationEngine/engine/commit/9adf3f33cf807f2cedafb227eb112c0faf10b0c6

This commit added `disableSigning` as well as `cert` and `keystore_property_file`

.sdkmanrc
NicoPiel: I agree. Removed.

settings.gradle.kts
```diff
@@ -0,0 +1,19 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ * For more detailed information on multi-project builds, please refer to https://docs.gradle.org/8.14/userguide/multi_project_builds.html in the Gradle documentation.
```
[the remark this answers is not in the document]
NicoPiel: It does exist.

tonygermano:
closing as no longer relevant

PR #223, opened 2025-12-10
refactor: Remove Xerces/xml-apis dependencies and modernize SAX readers

tonygermano:
Removes the dependency on `xercesImpl` and `xml-apis` to resolve classloader conflicts and reduce security risks in Java 17+ environments. Refactors custom Reader classes to use the standard JDK `XMLReader` interface instead of extending the Apache Xerces `SAXParser`.

Changes:
- **Dependency Removal:** Removed `xercesImpl` and `xml-apis` from the build configuration.
- **New Base Class:** Introduced `AbstractXMLReader` to centralize common `XMLReader` logic (handler management, feature/property compliance) and reduce code duplication across plugins.
- **Reader Refactoring:** Updated `DelimitedReader`, `EDIReader`, `ER7Reader`, and `NCPDPReader`:
  - Replaced inheritance (`extends SAXParser`) with `extends AbstractXMLReader`.
  - Added standard `parse(String)` overload via the base class.
  - Implemented strict SAX2 compliance for `getFeature`/`setFeature` to correctly report namespace support to downstream transformers.
- **Code Cleanup:** Removed unused imports of `org.apache.xerces.parsers.SAXParser`.
- **Documentation Cleanup:** Removed references to removed libraries in `THIRD-PARTY-README.txt`

This change ensures the application uses the built-in JDK XML parser, reducing the artifact size and aligning with modern Java best practices.

### DIFF
--- a/server/src/com/mirth/connect/plugins/datatypes/hl7v2/ER7Reader.java
+++ b/server/src/com/mirth/connect/plugins/datatypes/hl7v2/ER7Reader.java
@@ -130,10 +130,10 @@ public class ER7Reader extends AbstractXMLReader {
 
                 if (segmentIndex == 0) {
                     documentHead = MESSAGE_ROOT_ID;
-                    contentHandler.startElement("", documentHead, "", null);
+                    contentHandler.startElement("", documentHead, "", getEmptyAttributes());
                 }
 
-                contentHandler.startElement("", segmentId, "", null);
+                contentHandler.startElement("", segmentId, "", getEmptyAttributes());
                 handleFieldOrRepetitions(contentHandler, fieldSeparator, componentSeparator, subcomponentSeparator, repetitionSeparator, escapeCharacter, segmentId, fieldTokenizer);
                 contentHandler.endElement("", segmentId, "");
             } else {
@@ -159,7 +159,7 @@ public class ER7Reader extends AbstractXMLReader {
             // the naming is SEG.<field#>
             if (field.equals(fieldSeparator)) {
                 if (atLastField) {
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
                     contentHandler.endElement("", segmentId + "." + fieldId, "");
                 }
 
@@ -175,11 +175,11 @@ public class ER7Reader extends AbstractXMLReader {
                 }
 
                 if (enteredHeader && (fieldId == 1)) {
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
                     contentHandler.characters(fieldSeparator.toCharArray(), 0, 1);
-                    contentHandler.endElement("", segmentId + "." + (fieldId), null);
+                    contentHandler.endElement("", segmentId + "." + (fieldId), "");
                     fieldId++;
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
 
                     char[] specialCharacters;
                     if (!subcomponentSeparator.isEmpty()) {
@@ -195,7 +195,7 @@ public class ER7Reader extends AbstractXMLReader {
                     }
 
                     contentHandler.characters(specialCharacters, 0, specialCharacters.length);
-                    contentHandler.endElement("", segmentId + "." + (fieldId), null);
+                    contentHandler.endElement("", segmentId + "." + (fieldId), "");
                 } else if (enteredHeader && (fieldId == 2)) {
                     // do nothing
                 } else {
@@ -209,7 +209,7 @@ public class ER7Reader extends AbstractXMLReader {
         }
 
         if (atLastField) {
-            contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+            contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
             contentHandler.endElement("", segmentId + "." + fieldId, "");
         }
     }
@@ -224,7 +224,7 @@ public class ER7Reader extends AbstractXMLReader {
             if (field.equals(repetitionSeparator)) {
                 // check for ~~
                 if (atLastRepetition) {
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
                     contentHandler.characters(EMPTY_CHAR_ARRAY, 0, 0);
                     contentHandler.endElement("", segmentId + "." + fieldId, "");
                 }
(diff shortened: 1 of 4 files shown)


kayyagari (approved):
Built with Java 17 and tested using a channel with both inbound and outbound datatypes set to a combination of Hl7v2.x and XML.

server/src/org/openintegrationengine/engine/plugins/datatypes/AbstractXMLReader.java
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @param name The property name, which is a fully-qualified URI.
+     * @return The current value of the property.
+     * @throws SAXNotRecognizedException If the property value can't be assigned or retrieved.
+     * @throws SAXNotSupportedException When the property name is recognized but its
+     *                                   value cannot be determined at this time.
+     */
+    @Override
+    public Object getProperty(String name) throws SAXNotRecognizedException, SAXNotSupportedException {
+        throw new SAXNotRecognizedException("Property not recognized: " + name);
```
NicoPiel, line 259: Some transformer pipelines probe standard SAX properties (e.g., lexical handler). Should we recognize a minimal set to reduce compatibility risk, even if unsupported?

server/src/org/openintegrationengine/engine/plugins/datatypes/AbstractXMLReader.java
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @param name The property name, which is a fully-qualified URI.
+     * @param value The requested value for the property.
+     * @throws SAXNotRecognizedException If the property value can't be assigned or retrieved.
+     * @throws SAXNotSupportedException When the property name is recognized but its
+     *                                   value cannot be set to the requested value.
+     */
+    @Override
+    public void setProperty(String name, Object value) throws SAXNotRecognizedException, SAXNotSupportedException {
+        throw new SAXNotRecognizedException("Property not recognized: " + name);
```
NicoPiel, line 273: Some transformer pipelines probe standard SAX properties (e.g., lexical handler). Should we recognize a minimal set to reduce compatibility risk, even if unsupported?

NicoPiel (commented):
Since the motivation includes “correctly report namespace support to downstream transformers,” can we add a regression test that runs a reader through the actual transformation path (SAXSource/Transformer) to confirm no additional feature/property probes break execution?

server/src/org/openintegrationengine/engine/plugins/datatypes/AbstractXMLReader.java
tonygermano: According to [LexicalHandler](https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/org/xml/sax/ext/LexicalHandler.html), support for this is optional.

> If the reader does not report lexical events, it will throw a [SAXNotRecognizedException](https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/org/xml/sax/SAXNotRecognizedException.html) when you attempt to register the handler.

server/src/org/openintegrationengine/engine/plugins/datatypes/AbstractXMLReader.java
tonygermano: According to [LexicalHandler](https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/org/xml/sax/ext/LexicalHandler.html), support for this is optional.

> If the reader does not report lexical events, it will throw a [SAXNotRecognizedException](https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/org/xml/sax/SAXNotRecognizedException.html) when you attempt to register the handler.

tonygermano:
@NicoPiel 

> Since the motivation includes “correctly report namespace support to downstream transformers,” can we add a regression test that runs a reader through the actual transformation path (SAXSource/Transformer) to confirm no additional feature/property probes break execution?

"correctly report namespace support to downstream transformers" is alluding to the fact that the [XMLReader.setFeature](https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/org/xml/sax/XMLReader.html#setFeature(java.lang.String,boolean)) method states:
> All XMLReaders are required to support setting http://xml.org/sax/features/namespaces to true and http://xml.org/sax/features/namespace-prefixes to false.

These features affect the argument values used when the `XMLReader` calls `ContentHandler.startElement` and `ContentHandler.endElement`. None of the plugins that extend this class actually read XML, but read other file types and convert them to xml, so nearly all xml features and properties will be unsupported. The existing plugins were hard coded (for the most part) to respond as if the two required features are in their minimum required state. Subclasses that wish to add support for additional features or properties can override the methods in this abstract class.

The main purpose of this PR was to remove the imports of `org.apache.xerces.parsers.SAXParser`. The proper way to do this seemed to be to implement `XMLReader` instead of extending `org.apache.xerces.parsers.SAXParser`. I created the abstract class to avoid having the same boilerplate implementations in all four plugins.

> can we add a regression test that runs a reader through the actual transformation path (SAXSource/Transformer) to confirm no additional feature/property probes break execution?

I pushed some additional updates that now include tests for the new class and some additional changes after reading more of the interface javadocs to make sure the plugins follow the contract (mainly passing an empty `Attributes` object instead of `null`.) I think the current behavior of throwing exceptions for any probes for unsupported options is correct per the spec, and to pretend that we support a feature/property that we don't support could be worse for compatibility.

When merging I plan to squash the first 4 commits and keep the last one separate.

server/src/com/mirth/connect/plugins/datatypes/hl7v2/ER7Reader.java
```diff
@@ -176,11 +175,11 @@ private void handleFieldOrRepetitions(ContentHandler contentHandler, String fiel
...
 
                 if (enteredHeader && (fieldId == 1)) {
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
                     contentHandler.characters(fieldSeparator.toCharArray(), 0, 1);
-                    contentHandler.endElement("", segmentId + "." + (fieldId), null);
+                    contentHandler.endElement("", segmentId + "." + (fieldId), "");
                     fieldId++;
-                    contentHandler.startElement("", segmentId + "." + fieldId, "", null);
+                    contentHandler.startElement("", segmentId + "." + fieldId, "", getEmptyAttributes());
```
jonbartels, line 182: Is there any benefit to DRYing up this repeated string pattern? It, or something similar, is in several files

server/src/com/mirth/connect/plugins/datatypes/ncpdp/NCPDPReader.java
```diff
@@ -16,15 +11,15 @@
...
 import org.xml.sax.ContentHandler;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.helpers.AttributesImpl;
 
-public class NCPDPReader extends SAXParser {
+public class NCPDPReader extends AbstractXMLReader {
     private Logger logger = LogManager.getLogger(this.getClass());
-
+
```
jonbartels, line 22: Stray whitespace. We need a linter once we get Gradleized

tonygermano: I really thought I caught all of those

server/src/com/mirth/connect/plugins/datatypes/hl7v2/ER7Reader.java
tonygermano: I dunno, maybe? Might be worth looking at for another PR.

The first and third arguments when calling startElement or endElement for these particular plugins will always be empty strings because they are related to namespaces, which are always empty for these implementations. The second arg is the localName. The 4th arg of startElement is usually, but not always an empty attributes object.

kayyagari (approved):

PR #232, opened 2025-12-24
AutoClosable Client

tonygermano:
Commit 1
---
    Make several private fields final in Client
    
    These fields are all either initialized in their declarations or in the
    constructor and then never reassigned.

Commit 2
---
    Make Client AutoCloseable and close() idempotent
    
    - Make AutoClosable so it can be used in try-with-resources
    - Removed null check for serverConnection since it cannot be null
    - Do not shutdown serverConnection or close JAX-RS client if this client
      is already closed.

### DIFF
--- a/server/src/com/mirth/connect/client/core/Client.java
+++ b/server/src/com/mirth/connect/client/core/Client.java
@@ -131,7 +131,11 @@ import com.mirth.connect.util.MirthSSLUtil;
 import com.mirth.connect.util.messagewriter.EncryptionType;
 import com.mirth.connect.util.messagewriter.MessageWriterOptions;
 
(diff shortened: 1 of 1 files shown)


server/src/com/mirth/connect/client/core/Client.java
```diff
@@ -2675,4 +2679,4 @@ public void setPluginProperties(String extensionName, Properties properties, boo
     public String getProperty(String group, String name) throws ClientException {
         return getServlet(ConfigurationServletInterface.class).getProperty(group, name);
     }
-}
\ No newline at end of file
+}
```
NicoPiel, line 2682: Newline deleted on purpose?

mgaffigan:
I love boring refactors!

server/src/com/mirth/connect/client/core/Client.java
tonygermano: I think I added one rather than deleting it. My editors are configured to always add the new line before EOF, and I usually just leave that in the commit if it needs to be added since the last line is usually just a closing curly brace, and it's easily observable that there are no code changes happening. I'm not sure why github isn't showing the 🚫 on either side.

NicoPiel: I like an extra line before EOF better anyway! :)

kayyagari (approved):
Yay, "I reviewed it' ;).

jonbartels (approved):
LGTM #modernJava

PR #234, opened 2026-01-02
Steps towards reproducible builds

tonygermano:
I recommend looking at the commits individually. See commit messages for details.

After these changes, the only files that remain different in the setup folder between two consecutive builds are `mirth-client-core.jar`, due to the `version.properties` file which still contains the build timestamp. I did not want to tackle that one in this same PR since it will require code changes, as that value is used in a ConfigurationController API method.

Comparing the output of two builds can then be performed as in the example below, and should take no more than a few seconds to run:
```bash
$ diff -rq setup/ ../tmp/setup/
Files setup/client-lib/mirth-client-core.jar and ../tmp/setup/client-lib/mirth-client-core.jar differ
Files setup/cli-lib/mirth-client-core.jar and ../tmp/setup/cli-lib/mirth-client-core.jar differ
Files setup/manager-lib/mirth-client-core.jar and ../tmp/setup/manager-lib/mirth-client-core.jar differ
Files setup/server-lib/mirth-client-core.jar and ../tmp/setup/server-lib/mirth-client-core.jar differ
$
```

_Edit: Here are the commit messages for each commit after the most recent push_

commit 7dc8a4f771aa8177e22b7eb8f12234ec8b39946f

    Build: Change some mirth references to OIE in ant files
    
commit e6868ea0b63903d7260dd8b61d93a0b95038d871

    Build: add property for jar entry modify times
    
    Previously all class files and resources were added to the built jars
    with the build time. With this change the jar, war, and zip tasks will
    use the time stamp from the `archive.entry.date` property as the
    modified times of the entries in order to work towards reproducable
    builds.
    
    This property is required and has been set in the build.properties
    files.

commit 21b56dee3681344725ef435ef33f572ce994dc18

    Build: remove timestamp from javadoc pages
    
    By default, the javadoc task inserts the current time into a comment in
    the javadoc pages, which is at odds with reproducible builds. This
    commit
    
    - sets the `-notimestamp` option to remove that timestamp
    - removes the `-html5` option which has no function in current javadoc
      versions. The output is always in html5.
    - adds an MPLv2 notification to the bottom of every page
    - changes the arg types from `line` to `value`. `line` implies
      shell-like parsing that splits arguments on spaces, but that is not
      how they are being used here.

### DIFF
--- a/server/build.xml
+++ b/server/build.xml
@@ -1203,9 +1203,12 @@
 		
 		<javadoc sourcepath="${src}:${jdk_src_path}" destdir="${docs.javadocs.userapi}" packagenames="com.mirth.connect.server.userutil,com.mirth.connect.userutil,com.mirth.connect.plugins.httpauth.userutil" link="https://docs.oracle.com/en/java/javase/17/docs/api/">
 			<classpath refid="classpath" />
-			<arg line="-html5"/>
-			<arg line="--add-modules=java.sql.rowset"/>
-			<arg line="--add-exports=java.sql.rowset/com.sun.rowset=ALL-UNNAMED"/>
+			<bottom>
+				<![CDATA[Licensed under the <a href="https://www.mozilla.org/MPL/2.0/">Mozilla Public License 2.0</a>.]]>
+			</bottom>
+			<arg value="-notimestamp"/>
+			<arg value="--add-modules=java.sql.rowset"/>
+			<arg value="--add-exports=java.sql.rowset/com.sun.rowset=ALL-UNNAMED"/>
 		</javadoc>
 	</target>
 


tonygermano:
> Can we get a more obviously "false" value? 2000-01-01 00:00:00 or 2099-12-31 00:00:00 come to mind.
> 
> Git seems a safe build dependency, and handling its absence seems likely to silently break over time. Let's make it required, with a fallback that the values can be passed explicitly. (Call git if the value is not set, but an error is an error).

With this PR, normally the first thing the build echoes out looks like this:
```
     [echo] Using git.commit.timestamp of [2026-01-06T07:13:14Z]
     [echo] Using git.commit.hash of [c9ae71a776d25857634bf80028c276593736661f]
     [echo] Using git.commit.name of [v4.5.2-52-gc9ae71a]
```

When the git calls fail it looks like:
```
     [exec] Result: 1
     [echo] Using git.commit.timestamp of [2025-03-25T22:28:08.664Z]
     [exec] Result: 1
     [echo] Using git.commit.hash of []
     [exec] Result: 1
     [echo] Using git.commit.name of [unavailable]
```

If git isn't present (like if someone unpacked a sources.tar.gz,) then it makes sense to default back to some other values to allow the build to continue without failing, because these values are only metadata. Originally, I had the default timestamp set to unix epoch, but it turns out that the zip structure is based on the 16-bit FAT filesystem, and doesn't support dates that early. Since I was picking an arbitrary date, I made it related to the project, but it's not necessarily a failure condition (I intentionally set `failonerror="false"`.) I figured if anyone was checking the date closely, cared about the value, and missed the build output above, a date from March 2025 would still throw up some flags considering it is now 2026. 

I pushed a change that moves the default values for when `git` is unavailable to the `build.properties` files instead of hard coding them in the `build.xml` files. For now, I don't think we need to consider `git` as a requirement to build. If something changes down the road where it becomes an actual requirement I will reconsider my opinion on hard failing when it's not available.

Since ant properties are immutable, the actual properties (`git.commit.timestamp` instead of `git.commit.timestamp.default`) can be added to `build.properties` or passed like `-Dgit.commit.timestamp=2000-01-01T00:00:00Z` to override what `git` would set, even when it is available.

The description of this PR has been edited to include all of the commit messages for the original and additional commits.

tonygermano:
> I have a few questions.
> 
> * Does this follow typical norms for reproducible builds? eg "is there some ant plugin that does this for us" _OR_ are we following some common patterns?
> * Is this work translatable to gradle? Is this adding a feature to the builds that will be easy to reproduce in other build tools? Ideally, is this something that other tools do for us?
> 
> The code appears to handle the described cases.

The things that are different between two builds from the same commit that this PR addresses:
1. **The modified dates of the files *inside* of the jar files.** Currently it uses the modified time from the .class files. This PR explicitly sets the value. Now I'm considering just killing the git feature that uses the timestamp from the commit, and always using a static value (still controlled by a property,) because I highly doubt anyone is unpacking jar files to check the timestamps on the files they contain. The date I had originally chosen was supposed to be somewhat of an easter egg (the timestamp when openintegrationengine.org was registered,) and it conveniently occurred between mirth 4.5.2 and the first OIE commit. The jar files themselves will still have the timestamp of when they were built. This is a feature of the ant zip, jar, and war tasks, and I assume gradle has a similar option on its zip-like tasks, which is probably easier to apply than in ant.
2. **A build timestamp hidden in a comment in the javadoc pages.** Removing this is an argument of the `javadoc` command. Likewise, the ant javadoc task has support for removing it, and I expect gradle javadoc task does as well. Additionally (screenshot above,) I was adding bottom text to the javadoc pages noting the MPL license and displaying the git commit object's friendly name from which it was built and linking it to the commit in github (which will give a 404 if the commit has not been pushed to github.) I'm thinking about killing that git feature, too.

I already mentioned in the PR description that this does not address the timestamp in the `version.properties` files. My hope is to replace that timestamp with the same git commit object name that this PR currently is displaying in the javadoc. That seems to be the most commonly used value as a reproducible build version that tells you exactly which git commit it came from as well as the most recent tag for both releases and snapshot builds.

The changes to the github action were made to support getting that value in the most efficient way possible. It downloads the commits and tags without downloading the associated trees and blobs (directories and file contents.) Any other method of getting the previous tags, I think, downloads more data than what I used. Without the tags, `git describe` doesn't work.

server/build.xml
```diff
@@ -1065,7 +1068,7 @@
 			<else>
 				<!-- modify jar manifests -->
 				<!-- don't modify bcprov, since it already has the manifest changes and is signed by BC -->
-				<echo message="[Thread Count: ${manifest_thread_count}] Modifying jar manifests to add Permissions: all-permissions; Codebase: *; Application-Name: Mirth Connect" />
+				<echo message="[Thread Count: ${manifest_thread_count}] Modifying jar manifests to add Permissions: all-permissions; Codebase: *; Application-Name: Open Integration Engine" />
```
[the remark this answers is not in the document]
tonygermano: I acknowledge the inconsistencies. It's better to not say "mirth" where not required, especially in user-facing messages.

webadmin/build.xml
```diff
@@ -1,5 +1,8 @@
-<project name="webadmin" basedir="." default="dist">
-	<!-- Initialize these values in case the build is not being called form the main Mirth Connect build -->
+<project name="webadmin" basedir="." default="dist" xmlns:unless="ant:unless">
+	<fail message="Required property archive.entry.date is not set"
+		unless:set="archive.entry.date" />
```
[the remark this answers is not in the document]
tonygermano: webadmin was already structured in this way, and we don't build it, anyway. I was just keeping it up to date.

webadmin/build.xml
```diff
@@ -1,5 +1,8 @@
...
+
+	<!-- Initialize these values in case the build is not being called form the main OIE build -->
```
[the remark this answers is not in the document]
tonygermano: Naming inconsistencies and spelling error already called out in other copilot comments. Not fixing since we do not build webadmin, and I was only keeping it up to date with other changes.

client/ant-build.xml
```diff
@@ -113,203 +116,203 @@
...
 
 		<mkdir dir="${plugins.imageviewer}" />
-		<jar destfile="${plugins.imageviewer}/imageviewer-client.jar" basedir="${classes}">
+		<jar destfile="${plugins.imageviewer}/imageviewer-client.jar" basedir="${classes}" modificationTime="${archive.entry.date}">
 			<include name="com/mirth/connect/plugins/imageviewer/**" />
 		</jar>
 
 		<mkdir dir="${plugins.javascriptrule}" />
-		<jar destfile="${plugins.javascriptrule}/javascriptrule-client.jar" basedir="${classes}">
+		<jar destfile="${plugins.javascriptrule}/javascriptrule-client.jar" basedir="${classes}" modificationTime="${archive.entry.date}">
```
jonbartels, line 255: Any way to DRY this up? Is there a regex or rule you used to make sure you got the `modificationTime` appended to all the `jar` elements?

client/build.properties
```diff
@@ -11,4 +11,7 @@ logs=${basedir}/logs
 dist=${basedir}/dist
 
 # jars
-client.jar=mirth-client.jar
\ No newline at end of file
+client.jar=mirth-client.jar
```
jonbartels, line 14: Why are line ending characters visible here? Is it problematic?

client/ant-build.xml
tonygermano: In gradle, I think it is much easier to add a configuration option to all zip/jar tasks in one place, but not that I know of in ant. I know I got them all because when you compare two builds there are no differences in these jar files. If I had missed one they would show up in the diff.

client/build.properties
tonygermano: I believe github is confused because line 14 had previously been the last line in the file, and it didn't have a trailing new line. I just double checked, and the entire file has windows line endings before and after my changes. I'm not sure why it's not showing the `^M` on line 13 along with the others, because it is there in the file.

PR #235, opened 2026-01-03
Adds OpenAPI oneOf for connector properties at runtime

NicoPiel:
Adds an OpenAPI/Swagger Schema annotation to the connector properties field to declare concrete property subclasses via oneOf. This enables correct polymorphic schema generation for connector properties, improving API documentation and client code generation.

Also imports the Schema annotation required for the new metadata.

### DIFF
--- a/server/src/com/mirth/connect/server/servlets/SwaggerServlet.java
+++ b/server/src/com/mirth/connect/server/servlets/SwaggerServlet.java
@@ -39,6 +39,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.reflections.Reflections;
 import org.reflections.scanners.SubTypesScanner;
+import org.reflections.util.ClasspathHelper;
 import org.reflections.util.ConfigurationBuilder;
 
 import com.mirth.connect.client.core.Version;
@@ -130,7 +131,7 @@ public class SwaggerServlet extends HttpServlet {
 
 	private SortedSet<Class<? extends ConnectorProperties>> findConnectorPropertiesSubtypes() {
 		Reflections reflections = new Reflections(new ConfigurationBuilder()
-				.forPackages("com.mirth.connect")
+				.addUrls(ClasspathHelper.forClassLoader())
 				.addScanners(new SubTypesScanner(false)));
 
 		SortedSet<Class<? extends ConnectorProperties>> subtypes = new TreeSet<>(


NicoPiel:
Previously missing connectors in a generated `index.d.ts` like `HttpDispatcherProperties` now show up.

mgaffigan (changes requested):
I think the intent is good, but this creates a compile-time dependency loop:

```mermaid
graph TD;
    com.mirth.connect.connectors.file-->core.mirth.connect.model
    core.mirth.connect.model-->com.mirth.connect.connectors.file
```

core.mirth.connect.model should be happily ignorant of implementations.  Any discovery of implementations would need to happen at runtime, but would be instance specific.

server/src/com/mirth/connect/server/servlets/SwaggerServlet.java
```diff
@@ -81,4 +95,78 @@ public void init(ServletConfig config) throws ServletException {
...
+		for (Class<? extends ConnectorProperties> subtype : subtypes) {
+			addSubtypeSchema(schemas, connectorPropertiesSchema, subtype);
+		}
+
+		schemas.put("ConnectorProperties", connectorPropertiesSchema);
+	}
+
+	private SortedSet<Class<? extends ConnectorProperties>> findConnectorPropertiesSubtypes() {
+		Reflections reflections = new Reflections(new ConfigurationBuilder()
+				.forPackages("com.mirth.connect")
```
tonygermano: question: Is it possible to scan the entire classpath?

I don't think you can assume that 3rd-party connectors will use this package name (or the connectors in this repo if we eventually migrate them to other packages.)

server/src/com/mirth/connect/model/Connector.java
tonygermano, line 1: Can you remove this file from the PR? I think you did not mean for the remaining two changes to still be in place.

tonygermano (changes requested):

server/src/com/mirth/connect/server/servlets/SwaggerServlet.java
NicoPiel: Yeah, I'm pretty sure that's entirely possible. I included that line for testing purposes.

NicoPiel: Changed.

mgaffigan (changes requested):
Blocking issues:
1. `getSimpleName()` is not unique.  Deserialization in the current disgusting deserializer is based on full name.
2. Is it proper to use `ModelConverters.getInstance()` directly?  It seems odd to me to assume the process will contain exactly one mapping of types to OpenAPI schema.  I would assume it is possible to have two differently configured model converters per process.

Non-blocking issues:
3. I do not like relying upon classpath scanning since it pushes us farther away from modern practices like DI and being compatible with AOT.  ConnectorProperties subtypes should be exposed by the plugin registry itself.
4. I assume that there are more polymorphic types than just `ConnectorProperties` - this does not seem structured to accommodate those.

Future work:
5. This does not adjust the deserializer itself (leaving JSON unsupported/nonmatching to the schema, and likely continuing to support RCE vectors).  Long term, we should be adding the `JsonTypeInfo` and dynamic equivalent of `JsonSubTypes`.  `ObjectMapper.registerSubtypes` seems relevant.  (see `allowTypes` for [some nauseous code](https://github.com/nextgenhealthcare/connect/blob/be90435c/core-util/src/com/mirth/connect/model/converters/ObjectXMLSerializer.java#L230-L258))

PR #239, opened 2026-01-14
Include tls-manager-plugin in-tree

mgaffigan:
Opening for discussion.  Incorporates https://github.com/NovaMap-Health/tls-manager-plugin via `git subtree` and an ant build shim.

Open issues:
- [ ] Windows build support
- [ ] .sdkman updates?
- [ ] Propagate "disableSigning" and "disableTests" / how are tests handled?
- [ ] Test build failures
- [ ] Use local jars instead of kaur's packages from maven central
- [ ] Install4J impacts
- [ ] The rest of the owl


kryskool:
Hi @mgaffigan 

Why use subtree here, instead of submodule ?

mgaffigan:
@kryskool 
> Why use subtree here, instead of submodule ?

It's been a few months, so my memory might be failing me, but I think it was because:

- submodule requires secondary steps or configuration after checkout (git clone is insufficient)
- submodule requires patches upstream / subtree allows in tree patches

PR #241, opened 2026-01-19
Fix information disclosure when lockout enabled

### LINKED ISSUE #240: [BUG] Username enumeration possible if lockout enabled
## **Describe the bug**

The `/users/_login` endpoint reveals whether a username exists in the system by returning different error messages and statuses based on the account's state. Specifically, when password retry limits are enabled, the API response for a valid username includes a "login attempts remaining" counter or a `FAIL_LOCKED_OUT` status. A non-existent username returns a generic failure. This allows an attacker to distinguish between valid and invalid usernames (Account Enumeration).

## **To Reproduce**

1. Enable account lockout features in Mirth (e.g., set `_MP_PASSWORD_RETRYLIMIT` to 5).
2. Attempt a login with a **non-existent** username (e.g., `notARealUser`).
* **Response:** Returns generic `FAIL` with "Incorrect username or password."

3. Attempt a login with a **known valid** username (e.g., `admin`) and an incorrect password.
* **Response:** Returns `FAIL` but includes specific metadata: `"2 login attempt(s) remaining for "admin"..."`

4. Continue failed attempts on the valid username until the limit is reached.
* **Response:** Returns `FAIL_LOCKED_OUT` and a message confirming the account is locked for a specific duration.

## **Expected behavior**

To prevent account enumeration, the server should return an identical response (Status Code, JSON Body, and Response Timing) regardless of whether the username exists, the password is incorrect, or the account is currently locked.

The recommended response is a generic:
`{"status": "FAIL", "message": "Invalid username or password"}`

## **Actual behavior**

The API discloses the existence of valid accounts through specific error strings and the `LoginStatus` enum:

* **Valid User/Bad Pass:** Discloses remaining attempts.
* **Locked Account:** Discloses `FAIL_LOCKED_OUT` and the time remaining.
* **Invalid User:** Returns a generic failure with no attempt counter.

## **Workaround(s)**

1. **Disable Lockout/Retry Limits:** Reverting to default settings (where these are off) stops the specific counter disclosure, though it increases the risk of brute-force success.
2. **Network Restrictions:** Do not expose the Mirth API/Administrator port to the public internet; restrict access to a management VPN or trusted IP range.  Use a WAF or other security appliance to limit username enumeration.
3. **External Auth:** Use a federated identity provider so Mirth is not handling the credential validation logic.

[Reported by Michael Hobbs](https://discord.com/channels/943670759891554316/1462545336819126456)

mgaffigan:
When lockout is enabled, enhanced response error messages allow an attacker to enumerate valid usernames.  Change adds new configuration option to disable username enumeration by returning the same error message for both invalid usernames and locked out accounts.

The option is enabled by default.  To revert, specify `password.allowEnumeration=true` in mirth.properties.

Closes #240

### DIFF
--- a/server/conf/mirth.properties
+++ b/server/conf/mirth.properties
@@ -20,6 +20,7 @@ password.expiration = 0
 password.graceperiod = 0
 password.reuseperiod = 0
 password.reuselimit = 0
+password.allowusernameenumeration = false
 
 # Only used for migration purposes, do not modify
 version = 4.5.2
--- a/server/src/com/mirth/connect/model/PasswordRequirements.java
+++ b/server/src/com/mirth/connect/model/PasswordRequirements.java
@@ -29,6 +29,7 @@ public class PasswordRequirements implements Serializable {
     private int gracePeriod;
     private int reusePeriod;
     private int reuseLimit;
+    private boolean allowUsernameEnumeration;
 
     public PasswordRequirements() {
         this.minLength = 0;
@@ -42,9 +43,18 @@ public class PasswordRequirements implements Serializable {
         this.gracePeriod = 0;
         this.reusePeriod = 0;
         this.reuseLimit = 0;
+        this.allowUsernameEnumeration = false;
     }
 
+    /**
+     * @deprecated Use {@link #PasswordRequirements(int, int, int, int, int, int, int, int, int, int, int, boolean)} instead.
+     */
+    @Deprecated
     public PasswordRequirements(int minLength, int minUpper, int minLower, int minNumeric, int minSpecial, int retryLimit, int lockoutPeriod, int expiration, int gracePeriod, int reusePeriod, int reuseLimit) {
+        this(minLength, minUpper, minLower, minNumeric, minSpecial, retryLimit, lockoutPeriod, expiration, gracePeriod, reusePeriod, reuseLimit, false);
+    }
+
+    public PasswordRequirements(int minLength, int minUpper, int minLower, int minNumeric, int minSpecial, int retryLimit, int lockoutPeriod, int expiration, int gracePeriod, int reusePeriod, int reuseLimit, boolean allowUsernameEnumeration) {
         this.minLength = minLength;
         this.minUpper = minUpper;
         this.minLower = minLower;
@@ -56,6 +66,7 @@ public class PasswordRequirements implements Serializable {
         this.gracePeriod = gracePeriod;
         this.reusePeriod = reusePeriod;
         this.reuseLimit = reuseLimit;
+        this.allowUsernameEnumeration = allowUsernameEnumeration;
     }
 
     public int getMinLength() {
@@ -144,5 +155,13 @@ public class PasswordRequirements implements Serializable {
 
     public void setReuseLimit(int reuseLimit) {
         this.reuseLimit = reuseLimit;
+    }
+
+    public boolean getAllowUsernameEnumeration() {
+        return allowUsernameEnumeration;
+    }
+
+    public void setAllowUsernameEnumeration(boolean allowUsernameEnumeration) {
+        this.allowUsernameEnumeration = allowUsernameEnumeration;
     }
 }
--- a/server/src/com/mirth/connect/server/controllers/DefaultUserController.java
+++ b/server/src/com/mirth/connect/server/controllers/DefaultUserController.java
@@ -43,6 +43,7 @@ import com.mirth.connect.server.util.StatementLock;
 public class DefaultUserController extends UserController {
     public static final String VACUUM_LOCK_PERSON_STATEMENT_ID = "User.vacuumPersonTable";
     public static final String VACUUM_LOCK_PREFERENCES_STATEMENT_ID = "User.vacuumPersonPreferencesTable";
+    private static final String INCORRECT_CREDENTIALS_MESSAGE = "Incorrect username or password.";
 
     private Logger logger = LogManager.getLogger(this.getClass());
     private ExtensionController extensionController = null;
@@ -292,6 +293,7 @@ public class DefaultUserController extends UserController {
             boolean authorized = false;
             Credentials credentials = null;
             LoginRequirementsChecker loginRequirementsChecker = null;
+            PasswordRequirements passwordRequirements = ControllerFactory.getFactory().createConfigurationController().getPasswordRequirements();
 
             // Retrieve the matching User
             User validUser = getUser(null, username);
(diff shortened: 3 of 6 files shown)


server/src/com/mirth/connect/server/util/PasswordRequirementsChecker.java
```diff
@@ -57,6 +57,7 @@ public class PasswordRequirementsChecker implements Serializable {
     private static final String PASSWORD_LOCKOUT_PERIOD = "password.lockoutperiod";
     private static final String PASSWORD_REUSE_PERIOD = "password.reuseperiod";
     private static final String PASSWORD_REUSE_LIMIT = "password.reuselimit";
+    private static final String PASSWORD_ALLOW_ENUMERATION = "password.allowEnumeration";
```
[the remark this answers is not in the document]
mgaffigan: Updated.

server/src/com/mirth/connect/server/controllers/DefaultUserController.java
```diff
@@ -300,7 +302,11 @@ public LoginStatus authorizeUser(String username, String plainPassword, String s
                 Digester digester = ControllerFactory.getFactory().createConfigurationController().getDigester();
                 loginRequirementsChecker = new LoginRequirementsChecker(validUser);
                 if (loginRequirementsChecker.isUserLockedOut()) {
-                    return new LoginStatus(LoginStatus.Status.FAIL_LOCKED_OUT, "User account \"" + username + "\" has been locked. You may attempt to login again in " + loginRequirementsChecker.getPrintableStrikeTimeRemaining() + ".");
+                    if (passwordRequirements.getAllowEnumeration()) {
+                        return new LoginStatus(LoginStatus.Status.FAIL_LOCKED_OUT, "User account \"" + username + "\" has been locked. You may attempt to login again in " + loginRequirementsChecker.getPrintableStrikeTimeRemaining() + ".");
+                    } else {
+                        return new LoginStatus(LoginStatus.Status.FAIL, INCORRECT_CREDENTIALS_MESSAGE);
```
tonygermano, line 308: question: Are timing attacks possible?

Since the goal of this PR is to prevent enumeration, does this early return (skipping the db credential lookup and digester match) open up the possibility of timing-based attacks? It is understood that you did not create this situation.

server/src/com/mirth/connect/model/PasswordRequirements.java
```diff
@@ -29,6 +29,7 @@ public class PasswordRequirements implements Serializable {
     private int gracePeriod;
     private int reusePeriod;
     private int reuseLimit;
+    private boolean allowEnumeration;
```
tonygermano: comment (non-blocking): yuck

I don't like that the property is added to this class, but I guess it's fine. This class already seems to be doing too much stuff not related to password requirements. The first 5 "min" properties all have to do with password complexity, the "reuse" ones still constitute what makes a valid password, but the rest of them are more related to login security policies. I suppose this new property fits in with the last group, but we might want to split them off at some point in the future.

server/src/com/mirth/connect/server/util/PasswordRequirementsChecker.java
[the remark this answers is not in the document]
tonygermano: Do you need to push? I don't see it updated.

server/src/com/mirth/connect/model/PasswordRequirements.java
tonygermano: suggestion: rename this property

Can we rename the property to something more descriptive of what it is doing, like `allowDetailedAuthErrors`?

We need to consider how it will be initialized if someone calls the non-default constructor. I think an uninitialized boolean defaults to false, but all other properties are explicitly set in both constructors.

tonygermano (changes requested):
suggestion: add default value to mirth.properties

I think the default should probably be added to the `/server/conf/mirth.properties` file as part of this change.

---
question (non-blocking): do we need to improve event logging with this change?

It kind of sucks prior to your changes when lockout is enabled. The event just shows the failed login attempts with the username as the only property. There's no indication to someone viewing the log that the account has become locked (you might be able to guess if you see repeated failed attempts.) Should we fire off another event when incrementing the strikes if it causes the account to lock? Putting the unlock time as a property of the server event would be helpful, I think.

This would also allow a future PR to send an email triggered by the event to the user notifying them of their account being locked now that it just appears they keep entering their credentials incorrectly.

Additionally, there doesn't appear to be a way for another admin to unlock it, and the lock duration is specified in hours.

server/src/com/mirth/connect/model/PasswordRequirements.java
mgaffigan: I've renamed it to `allowUsernameEnumeration`.  I am avoiding the specific implementation detail that is "allowing detailed auth errors", since this also should effect other things (e.g.: slow login with random noise, IP-based lockout, etc...).

I want the name to sound concerning (à la `dangerouslySetInnerHTML`).  `allowDetailedAuthErrors` sounds like something I should enable to be nice.  `allowUsernameEnumeration` sounds like an information disclosure vulnerability.

mgaffigan: I don't think the class has unrelated things in it - just a bad name.  It really ends up being the authentication system settings.  I don't think this is worth a breaking change to the API.

pacmano1: Whatever you all do here - make sure that should the value not be set at all in mirth.properties the current default behavior does not change or whatever make the most sense here.

The installer does not currently edit / inspect mirth.properites.

tonygermano: Agree with the potential for it to sound too nice. Originally I was going to suggest `suppressDetailedAuthErrors` but I didn't know if it would be less desirable to have a new setting that needed to default to true.

I don't know about having one generic setting that does a lot of stuff, though, because it would be unclear about what happens when you turn it on, or some people might want some functionality, but not all of it. I think a cautionary statement can be put in a comment right above the setting in mirth.properties and in any documentation.

I don't think anyone would turn on the setting to specifically allow username enumeration, but they might want detailed auth messages if they understand the repercussions of turning it on. E.g., if I can still prevent brute force attacks by banning IP addresses, or my instance is not even publicly accessible, I might want to let my users know when their account is locked or password is expired other than telling them they are typing their credentials incorrectly.

server/src/com/mirth/connect/server/controllers/DefaultUserController.java
mgaffigan: Re: Timing attacks, absolutely an issue.  I recommend a different issue and PR.

server/src/com/mirth/connect/model/PasswordRequirements.java
mgaffigan: "suppressDetailedAuthErrors" is not "scary", and a comment in mirth.properties does not extend to environment variables and the code itself.  I'm not planning any further edits at this time.  If this is blocking, please clarify.

server/conf/mirth.properties
```diff
@@ -20,6 +20,7 @@ password.expiration = 0
 password.graceperiod = 0
 password.reuseperiod = 0
 password.reuselimit = 0
+password.allowusernameenumeration = false
```
tonygermano, line 23: This is what I meant by naming the property with its function and including a warning about a potential consequence of changing the value rather than naming the property after the potential consequence and leaving people to wonder about what it actually does.

```suggestion
# Warning: Changing suppressdetailedautherrors to false could allow user enumeration by an attacker
password.suppressdetailedautherrors = true
```

I don't think we need to worry about not being able to put the warning on setting the values via environment variables, because the only way people would discover that the property exists in the first place is by looking at the mirth.properties file or reading external documentation about the properties, both of which would include the warning.

If you still want to keep the default value of false, then I think `allowdetailedautherrors` is already enough to give users pause. Using "allow" rather than another word like "show" already implies it is being intentionally withheld. Combined with the warning, it should be sufficiently scary.

If another aspect of preventing user enumeration requires an option in mirth.properties, rather than being always implemented, then I don't think it should share functionality with this property, but should be its own setting, also with an accurate description of what it is doing that a user might be concerned with besides preventing user enumeration. The desire to be scary can't mask the true functionality of the setting.

mgaffigan: I have no further edits planned to the name. If this is blocking, feel free to take ownership of the PR.

mgaffigan:
Replaced by #439

PR #243, opened 2026-01-21
Feature/modern build system (gradle without requiring ant)

jessesaga:
I see there are other open PRs related to gradle build approaches. I'm not sure if this one adds value above and beyond what others are working on but I figured I'd submit it just in case. 

This PR adds a native gradle build capabilities that fully builds tar.gz, deb, and rpm packages. There is also a github action that automates all of this.

There are probably still bugs and inadequacies with this PR but the validated bits include

- github action build, packaging, and release works (minus test cases currently)
- the generated tar.gz runs well on Linux and connecting via Mirth Administrator and various operations were verified

TODOs include

- The RPM and Deb packages have not been validated and there is no Windows build.
- Opening the project in eclipse or other IDEs has not been validated.
- This PR currently clobbers the existing github action and .classpath files

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -105,7 +105,7 @@ jobs:
           fi
 
       - name: Build with Gradle
-        run: ./gradlew build assembleSetup -x :donkey:test -x :server:test --no-daemon
+        run: ./gradlew build assembleSetup buildLinuxPackages -x :donkey:test -x :server:test --no-daemon
 
       - name: Run Donkey integration tests
         if: ${{ !inputs.skip_tests }}
@@ -119,9 +119,6 @@ jobs:
           name: donkey-test-results
           path: donkey/build/reports/tests/test/
           if-no-files-found: ignore
-
-      - name: Build Linux packages
-        run: ./gradlew buildLinuxPackages --no-daemon
 
       - name: List built packages
         run: |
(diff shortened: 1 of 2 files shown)


donkey/src/test/java/com/mirth/connect/donkey/test/util/TestUtils.java
```diff
@@ -1291,6 +1292,9 @@ public static DonkeyConfiguration getDonkeyTestConfiguration() {
             databaseProperties.load(is);
             IOUtils.closeQuietly(is);
 
+            // Initialize connection pools before returning configuration
```
NicoPiel: Can you explain why this is necessary? Does it help with tests?

donkey/src/test/java/com/mirth/connect/donkey/test/util/TestUtils.java
```diff
@@ -1291,6 +1292,9 @@ public static DonkeyConfiguration getDonkeyTestConfiguration() {
...
+            DonkeyConnectionPools.getInstance().init(databaseProperties);
```
NicoPiel, line 1296: Can you explain why this is necessary? Does it help with testing?

jessesaga: It's not necessary. This PR does not currently have test cases working fully. This was an attempt to start fixing that but is not relevant for the PR unless we include getting the test suite working.

NicoPiel: I see. In that case I think it would be better to exclude these changes from the PR, as it is already a massive thing to review.
Test cases have been a pain point with the other Gradle PRs as well, you're not alone.

mgaffigan:
Love to see it! I've not gotten to review, but look forward to doing so.

Current view is that we'll need to merge piecemeal, but if we can get to all passing and byte identical releases, I think a single PR might be possible.

donkey/src/test/java/com/mirth/connect/donkey/test/util/TestUtils.java
jessesaga: Sounds good. I'll remove it.

PR #249, opened 2026-02-12
Maint: Remove webadmin and manager sub-projects

tonygermano:
These are not used by OIE.

PR #252, opened 2026-02-16
maint: cleanup incomplete removal of webadmin and manager

tonygermano:
The `.project` files were missed in #249 . I'm not quite sure how, as I had completely removed the directories.


pacmano1 (changes requested):
{basedir}/manager/lib is still in dependency-check.xml.

tonygermano:
> {basedir}/manager/lib is still in dependency-check.xml.

@pacmano1 I don't think we have all of the pieces to make dependency-check.xml work, but I updated it, anyway.

kryskool (approved):

Issue #258, opened 2026-03-05
[IDEA] serverName not available on alert template


kryskool:
**Is your feature request related to a problem? Please describe.**
When we configure an alert, we have only serverId and not the name, it's difficult to see which server is relate when we receive an email

**Describe your use case**
send an alert email with server name instead of **serverId** only

**Describe the solution you'd like**
Add serverName as variable

**Describe alternatives you've considered**
N/A

**Additional context**

<img width="171" height="306" alt="Image" src="https://github.com/user-attachments/assets/cc4a1cac-21ed-4078-9750-f758bdab0533" />

PR #259, opened 2026-03-05
feat: add serverName variable on alert template

kryskool:
Ref #258

### DIFF
--- a/client/src/com/mirth/connect/client/ui/alert/DefaultAlertEditPanel.java
+++ b/client/src/com/mirth/connect/client/ui/alert/DefaultAlertEditPanel.java
@@ -105,6 +105,7 @@ public class DefaultAlertEditPanel extends AlertEditPanel {
         variables.add("alertId");
         variables.add("alertName");
         variables.add("serverId");
+        variables.add("serverName");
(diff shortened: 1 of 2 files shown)


tonygermano (approved):
So, I just realized when seeing the results of the co-pilot review Sean kicked off two days ago that this is actually a partial duplicate of #47 , which adds `serverName` and `environmentName`. It couldn't be merged at the time that it was opened because it required at least java 9, and we had not upgraded from java 8 to 17 yet.

In regard to the question I had asked above, I verified that the internal value can't actually be an empty string, only null. I tested this by
- starting a new instance
- setting values for both fields in the ui and saving
- checking the new values were being used
- clearing the values in the ui and saving
- verifying the internal values were set back to null rather than an empty string

Attempting to use null values in the template results in the token remaining in the output and not being replaced rather than showing an empty value.

PR #47 already addresses this by inserting an empty string into the alert velocity context for `serverName` or `environmentName` if either of them are null. As far as I can tell, that seems to be an acceptable solution for everyone who has already commented on this PR.

I'm going to merge this one as it is since most of the discussion has happened here, and then we'll get the other one applied on top of it.

PR #262, opened 2026-03-07
build: add AbstractXMLReader to mirth-client-core

tonygermano:
This class was not available to the client, but it is required to serialize transformer message templates to the message tree view.


pacmano1 (dismissed):
Tested with JSON, XML and HL7.   Seems fixed.  Excellent find whoever found it.

NicoPiel (dismissed):
Superb find.

PR #263, opened 2026-03-10
feat: add JMS template for RabbitMQ

kryskool:
Add template for JMS connection with RabbitMQ

SCR-20260313-inuz

### DIFF
--- a/client/src/com/mirth/connect/connectors/jms/JmsConnectorPanel.java
+++ b/client/src/com/mirth/connect/connectors/jms/JmsConnectorPanel.java
@@ -643,6 +643,10 @@ public class JmsConnectorPanel extends ConnectorSettingsPanel {
                 connectionFactoryNameField.setText(template.getJndiConnectionFactoryName());
                 connectionFactoryClassField.setText(template.getConnectionFactoryClass());
                 connectionPropertiesTable.setProperties(template.getConnectionProperties());
+                usernameField.setText(template.getUsername());
+                passwordField.setText(template.getPassword());
+                clientIdField.setText(template.getClientId());
+                destinationNameField.setText(template.getDestinationName());
             }
         }
     }//GEN-LAST:event_loadTemplateButtonActionPerformed


client/src/com/mirth/connect/connectors/jms/JmsConnectorPanel.java
```diff
@@ -643,6 +643,10 @@ private void loadTemplateButtonActionPerformed(java.awt.event.ActionEvent evt) {
                 connectionFactoryNameField.setText(template.getJndiConnectionFactoryName());
                 connectionFactoryClassField.setText(template.getConnectionFactoryClass());
                 connectionPropertiesTable.setProperties(template.getConnectionProperties());
+                usernameField.setText(template.getUsername());
+                passwordField.setText(template.getPassword());
+                clientIdField.setText(template.getClientId());
+                destinationNameField.setText(template.getDestinationName());
```
tonygermano: I don't think these fields were meant to be saved in a template. They are already clearly labeled in the GUI, are specific to a particular connection, and could contain sensitive data.

tonygermano, line 649:
```suggestion
```

I don't think these fields were meant to be saved in a template. They are already clearly labeled in the GUI, are specific to a particular connection, and could contain sensitive data.

client/src/com/mirth/connect/connectors/jms/JmsTemplateListModel.java
```diff
@@ -53,6 +53,17 @@ private JmsTemplateListModel() {
...
 
+        properties = new JmsConnectorProperties();
+        properties.setUseJndi(false);        
+        properties.setConnectionFactoryClass("com.rabbitmq.jms.admin.RMQConnectionFactory");
+        properties.getConnectionProperties().put("brokerURL", "failover:(tcp://localhost:5672)?maxReconnectAttempts=0");
+        properties.getConnectionProperties().put("closeTimeout", "15000");
+        properties.getConnectionProperties().put("useCompression", "no");
+        properties.setUsername("guest");
+        properties.setPassword("guest");
+        properties.setDestinationName("myqueue");
```
tonygermano, line 64:
```suggestion
        properties.getConnectionProperties().put("host", "localhost");
        properties.getConnectionProperties().put("port", "5672");
        properties.getConnectionProperties().put("virtualHost", "/");
```

Similar to my other comment, I don't think username, password, and destinationName are meant to be saved in the template.

I believe the connection properties are not valid RabbitMQ properties. They appear to be copied from the ActiveMQ template. `host`, `port`, and `virtualHost` are what appear in the [documentation](https://rabbitmq.github.io/rabbitmq-jms-client/2.x/stable/htmlsingle/index.html#defining-the-jms-connection-factory). I tested them with non-default values and they worked. You can see what the other valid properties are by scrolling down a bit in that section.

tonygermano (changes requested):
I left a couple suggestions for changes.

This connector could use some work in a separate PR. It was a pain to test. It only supports jms 1.1 by default. It supports jms 2 with an additional javax.jms-api library, but not jms 3 (jakarta.jms,) and the resource dir only worked when "Load parent-first" was selected.

PR #278, opened 2026-03-28
 Fix: Add JavaFX requirement check on startup

mgaffigan:
Adds a startup check to verify the environment is acceptable to the app.  Can't check for java version since we're targeting bytecode 61, but does check Java FX.

Here's the new startup error.

image

Tested using Azul Zulu JRE 17 FX and SE

Closes #277


mgaffigan:
@NicoPiel,

Currently, it only has the check for the JavaFX prerequisite:

> Open Integration Engine Client requires JavaFX.  Please relaunch with a supported JRE.

But it could be extended in the future for the JRE version:

> Open integration Engine Client requires JRE 24.  Please relaunch with a supported JRE.

jonbartels (approved):

NicoPiel (approved):
Sure, let's get it merged, then.

kryskool (approved):

kpalang (approved):

PR #279, opened 2026-03-29
Feature: Add recent filters drop-down to the message browser

### LINKED ISSUE #273: [IDEA] Recent filters drop down
Often the user uses the same filters over and over on the same channel.  Allow a MRU list of 10 or so filters to avoid having to specify each time.

mgaffigan:
Adds a recent button to the message browser pane.  Allows restoration of the last 10 searches on the channel.

image
image

Closes #273


client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
```diff
@@ -684,7 +763,7 @@ protected void runSearch() {
             clearCache();
             loadPageNumber(1);
 
-            updateSearchCriteriaPane();
+            lastSearchCriteria.setText(messageFilter.toString(connectors, "\n", /* includeEmptyCriteria: */ true));
```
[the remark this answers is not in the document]
mgaffigan: Intentional change: MessageFilter does not contain state to allow identification of "(all day)".

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -13,13 +13,21 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
 import com.mirth.connect.donkey.model.message.Status;
 import com.mirth.connect.model.MessageFilterToStringStyle;
```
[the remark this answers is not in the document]
mgaffigan: Reverting toString makes these used again.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,230 @@ public void setError(Boolean error) {
...
+        return this.equals(EMPTY_FILTER);
+    }
+
     @Override
     public String toString() {
-        return ToStringBuilder.reflectionToString(this, MessageFilterToStringStyle.instance());
+        return toString(Map.of(), "\n", false);
+    }
+
+    public String toString(Map<Integer, String> connectors, String padding, boolean includeEmptyCriteria) {
```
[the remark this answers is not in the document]
mgaffigan: Accepted.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java
```diff
@@ -452,6 +455,75 @@ public void updateAdvancedSearchButtonFont() {
...
+        // Multi-channel browse may be null
+        if (recentFilterStore == null) return;
+
+        JPopupMenu popupMenu = new JPopupMenu();
+        for (MessageFilter filter : recentFilterStore.getRecentFilters()) {
+            var menuItem = new JMenuItem(filter.toString(connectors, "; ", /* includeEmptyCriteria: */ false));
+            menuItem.addActionListener(e -> applyMessageFilter(filter));
+            popupMenu.add(menuItem);
+        }
+        popupMenu.show(recentFiltersButton, 0, recentFiltersButton.getHeight());
```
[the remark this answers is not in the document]
mgaffigan: Null list is accepted, but the rest seems extra.  We're fine to fail in showing the menu - we won't corrupt state.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,68 @@
...
+    public List<MessageFilter> getRecentFilters() {
+        try {
+            String serialized = Preferences.userNodeForPackage(Mirth.class).get(prefKey, "");
+            if (StringUtils.isBlank(serialized)) return List.of();
+
+            return (List<MessageFilter>) ObjectXMLSerializer.getInstance().deserialize(serialized, List.class);
+        } catch (Exception e) {
+            // Fail quietly if the stored filters cannot be deserialized for any reason.
+            e.printStackTrace();
+            return List.of();
```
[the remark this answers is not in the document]
mgaffigan: This is the first version.  That's future me's problem.

NicoPiel:
Great feature. Will look at the code as soon as I am able to.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.prefs.Preferences;
+
+import org.apache.commons.lang3.StringUtils;
+
+import com.mirth.connect.client.ui.Mirth;
+import com.mirth.connect.model.converters.ObjectXMLSerializer;
+import com.mirth.connect.model.filters.MessageFilter;
+
+class MessageBrowserRecentFilterStore {
+    private static final int MAX_RECENT_FILTERS = 10;
```
jonbartels, line 18: Max should be configurable in `mirth.properties`

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,71 @@
...
+        var filters = new ArrayList<MessageFilter>(getRecentFilters());
+
+        // Remove then re-add to avoid duplicates
+        filters.remove(filter);
+        filters.add(0, filter);
+
+        // Trim to the maximum number of recent filters
+        if (filters.size() > MAX_RECENT_FILTERS) {
+            filters.subList(MAX_RECENT_FILTERS, filters.size()).clear();
+        }
```
jonbartels, line 61: What about using https://guava.dev/releases/19.0/api/docs/com/google/common/collect/EvictingQueue.html or https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/queue/CircularFifoQueue.html here?

They should be available and provide the behavior you want

jonbartels (changes requested):
Pretty nice! Left two comments for improvements to property handling and maybe a simpler data structure

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Can we take a feature issue on this? It seems useful even with the MRU size hard coded.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: @jonbartels, I toyed around with this today, but using a "proper" queue has some negative effects.  Firstly: we're not strictly a circular buffer since we pop duplicates from the middle of the queue.  Secondly: we don't ever maintain the datastructure over time - the primary intent is to serialize it (which works better with a vanilla ArrayList since it produces a vanilla `<list>`).

Unless this is a blocker for you, I am calling it on this one.

jonbartels: Good enough for me. Tx for discussing!

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
jonbartels: Sure. How about a broad feature issue that finds hardcoded constants or limits in the Client and makes them configurable?

mgaffigan: That sounds like a big to me, but maybe with sub-issues for each constant, sure.  The issues I had in mind were:

1. Allow administrators to default and lock user preferences via mirth.properties
2. Make `MAX_RECENT_FILTERS` a user preference

pacmano1:
This feature may leave PII and PHI locally on a workstaiton.  So in its current storage format, it's a no for me.

pacmano1 (dismissed):
Can leave PII/PHI on local workstation.  Storage format needs to be changed.

gibson9583:
> This feature may leave PII and PHI locally on a workstaiton. So in its current storage format, it's a no for me.

agree that PII/PHI is a risk. Perhaps we should rethink how this works? While it may negate how useful this is, perhaps this is cleared on user logout? If thats the case, maybe we then dont cache by channelId but rather global so user doesnt have to build the cache on every channel.

pacmano1:
> > This feature may leave PII and PHI locally on a workstaiton. So in its current storage format, it's a no for me.
> 
> agree that PII/PHI is a risk. Perhaps we should rethink how this works? While it may negate how useful this is, perhaps this is cleared on user logout? If thats the case, maybe we then dont cache by channelId but rather global so user doesnt have to build the cache on every channel.

admin session can be terminated without logout.

mgaffigan:
@pacmano1 / @gibson9583 
> > This feature may leave PII and PHI locally on a workstaiton. So in its current storage format, it's a no for me.
> 
> agree that PII/PHI is a risk. Perhaps we should rethink how this works? While it may negate how useful this is, perhaps this is cleared on user logout? If thats the case, maybe we then dont cache by channelId but rather global so user doesnt have to build the cache on every channel.

I agree with the concern, but think storing any text search only in memory or in preferences is likely sufficient.  I'll update the PR to address the unencrypted-local-storage of text searches.

gibson9583:
> @pacmano1 / @gibson9583
> 
> > > This feature may leave PII and PHI locally on a workstaiton. So in its current storage format, it's a no for me.
> > 
> > 
> > agree that PII/PHI is a risk. Perhaps we should rethink how this works? While it may negate how useful this is, perhaps this is cleared on user logout? If thats the case, maybe we then dont cache by channelId but rather global so user doesnt have to build the cache on every channel.
> 
> I agree with the concern, but think storing any text search only in memory or in preferences is likely sufficient. I'll update the PR to address the unencrypted-local-storage of text searches.

in-memory is good with me. It doesnt need to survive client restart. Preferences is the only issue I can see because that is stored unencrypted and available to any process on server.

mgaffigan:
@pacmano1 / @gibson9583, updated to use in-memory storage only.  That's tantamount to the rest of the PHI which already resides in memory.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) Mirth Corporation. All rights reserved.
+ * 
+ * http://www.mirthcorp.com
+ * 
+ * The software in this package is published under the terms of the MPL license a copy of which has
+ * been included with this distribution in the LICENSE.txt file.
+ */
```
kpalang: question: Why Mirth Corporation copyright if this is a new class?

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,73 @@
...
+    private final String prefKey;
+
+    public MessageBrowserRecentFilterStore(String channelId) {
+        this.prefKey = RECENT_FILTERS_PREFERENCE_PREFIX + channelId;
+    }
+
+    public List<MessageFilter> getRecentFilters() {
+        try {
+            String serialized = RECENT_FILTERS_BY_CHANNEL.getOrDefault(prefKey, "");
+            if (StringUtils.isBlank(serialized)) return List.of();
```
kpalang: suggestion: Use stdlib's `String.isBlank()` instead of the library call.

The StringUtils from commons is typically used to check for `null` and whitespace in a single call, but here it's realistically impossible to get a `null` value, as the `prefKey` is either concatenated from a certainly non-null string or defaulted to an empty string.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,73 @@
...
+
+            var result = ObjectXMLSerializer.getInstance().deserialize(serialized, List.class);
+            if (result == null) return List.of();
+            
+            return (List<MessageFilter>) result;
```
kpalang: question: Would using the `ObjectXMLSerializer.getInstance().deserializeList(serialized, MessageFilter.class)` call solve the unchecked (and redundant) cast squiggly?

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,73 @@
...
+        } catch (Exception e) {
+            // Fail quietly if the stored filters cannot be deserialized for any reason.
+            e.printStackTrace();
```
kpalang: suggestion: I know you hate future you, but add proper logging and possibly prepopulate `log2j.xml` with the appropriate knobs for visibility control.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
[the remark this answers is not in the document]
kpalang: `new ArrayList<>()`

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
```diff
@@ -0,0 +1,73 @@
...
+            return List.of();
+        }
+    }
+
+    public void addRecentFilter(MessageFilter filter) {
+        if (filter == null) {
+            throw new IllegalArgumentException("Filter cannot be null");
+        }
+
+        var filters = new ArrayList<MessageFilter>(getRecentFilters());
```
kpalang: nitpick: Redundant explicit type argument.

server/src/com/mirth/connect/model/filters/elements/ContentSearchElement.java
```diff
@@ -41,4 +42,22 @@ public void setSearches(List<String> searches) {
     public String toString() {
         return ToStringBuilder.reflectionToString(this, ContentSearchElementToStringStyle.instance());
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof ContentSearchElement
+            && equals((ContentSearchElement) obj);
```
kpalang: suggestion: `obj instanceof ContentSearchElement cse` pattern matching because pretty.

server/src/com/mirth/connect/model/filters/elements/MetaDataSearchElement.java
```diff
@@ -61,4 +62,24 @@ public void setIgnoreCase(Boolean ignoreCase) {
     public String toString() {
         return ToStringBuilder.reflectionToString(this, SearchElementToStringStyle.instance());
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof MetaDataSearchElement
+            && equals((MetaDataSearchElement) obj);
```
kpalang: suggestion: `obj instanceof MetaDataSearchElement mdse` pattern matching because pretty.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
```diff
@@ -510,6 +510,81 @@ public Boolean hasAdvancedCriteria() {
...
+        }
+    }
+
+    private String formatMetaDataValue(Object value) {
+        if (value == null) {
+            return "";
+        }
+
+        if (value instanceof java.util.Calendar) {
+            return new java.text.SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(((java.util.Calendar) value).getTime());
```
kpalang: question: Why qualified names instead of imports?

https://i.kym-cdn.com/entries/icons/original/000/054/773/doakescover.jpg

kpalang: suggestion: `value instanceof Calendar calendar` pattern matching because pretty.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
```diff
@@ -510,6 +510,81 @@ public Boolean hasAdvancedCriteria() {
...
+
+        messageIdLowerField.setText((messageFilter.getMinMessageId() == null) ? "" : String.valueOf(messageFilter.getMinMessageId()));
+        messageIdUpperField.setText((messageFilter.getMaxMessageId() == null) ? "" : String.valueOf(messageFilter.getMaxMessageId()));
+        originalIdLowerField.setText((messageFilter.getOriginalIdLower() == null) ? "" : String.valueOf(messageFilter.getOriginalIdLower()));
+        originalIdUpperField.setText((messageFilter.getOriginalIdUpper() == null) ? "" : String.valueOf(messageFilter.getOriginalIdUpper()));
+        importIdLowerField.setText((messageFilter.getImportIdLower() == null) ? "" : String.valueOf(messageFilter.getImportIdLower()));
+        importIdUpperField.setText((messageFilter.getImportIdUpper() == null) ? "" : String.valueOf(messageFilter.getImportIdUpper()));
+        serverIdField.setText(StringUtils.defaultString(messageFilter.getServerId()));
+        sendAttemptsLower.setValue((messageFilter.getSendAttemptsLower() == null) ? 0 : messageFilter.getSendAttemptsLower());
+        sendAttemptsUpper.setValue((messageFilter.getSendAttemptsUpper() == null) ? "" : String.valueOf(messageFilter.getSendAttemptsUpper()));
```
kpalang: suggestion: I'd write a helper function for the value generation. The repetition is hard to read.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
         this.error = error;
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof MessageFilter
+            && equals((MessageFilter) obj);
```
kpalang: suggestion: Pattern matching because pretty.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
...
+    public String toDisplayString(Map<Integer, String> connectors, String padding, boolean includeEmptyCriteria) {
+        StringBuilder text = new StringBuilder();
+
+        if (maxMessageId != null) {
+            text.append("Max Message Id: ");
+            text.append(maxMessageId);
+        }
+
+        if (minMessageId != null) {
+            if (text.length() > 0) text.append(padding);
```
kpalang: question: Why the length check instead of `!isEmpty()`?

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
...
+            if (statuses == null) {
+                text.append("(any)");
+            } else {
+                text.append(StringUtils.join(statuses, ", "));
+            }
+        }
+
+        if (textSearch != null) {
+            if (text.length() > 0) text.append(padding);
+            text.append("Text Search: " + textSearch);
```
kpalang: nitpick: Intellij complains about a concat inside an `append()`.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
...
+
+        if (originalIdLower != null || originalIdUpper != null) {
+            if (text.length() > 0) text.append(padding);
+            text.append("Original Id: ");
+            if (originalIdUpper == null) {
+                text.append("Greater than " + originalIdLower);
+            } else if (originalIdLower == null) {
+                text.append("Less than " + originalIdUpper);
+            } else {
+                text.append("Between " + originalIdLower + " and " + originalIdUpper);
```
kpalang: suggestion: A `String.formatted()` would be more readable here.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
...
+                }
+            }
+        }
+
+        if (metaDataSearch != null) {
+            for (MetaDataSearchElement element : metaDataSearch) {
+                if (text.length() > 0) text.append(padding);
+                text.append(element.getColumnName() + " " + MetaDataSearchOperator.fromString(element.getOperator()).toString() + " ");
+                if (element.getValue() instanceof Calendar) {
+                    Calendar date = (Calendar) element.getValue();
```
kpalang: suggestion: Pattern matching because pretty.

server/src/com/mirth/connect/model/filters/MessageFilter.java
```diff
@@ -222,8 +236,244 @@ public void setError(Boolean error) {
...
+        }
+
+        if (textSearch != null) {
+            if (text.length() > 0) text.append(padding);
+            text.append("Text Search: " + textSearch);
+        }
+
+        if (includeEmptyCriteria
+            || (includedMetaDataIds != null && !includedMetaDataIds.isEmpty())
+            || (excludedMetaDataIds != null && !excludedMetaDataIds.isEmpty())) {
```
kpalang: nitpick: I prefer having the `}` at on a new line to create a better visual separation.

```java
if (includeEmptyCriteria
    || (includedMetaDataIds != null && !includedMetaDataIds.isEmpty())
    || (excludedMetaDataIds != null && !excludedMetaDataIds.isEmpty()))
{
    getConnectorSearchCriteriaText(text, padding, connectors);
}
```

kpalang (changes requested):
Some suggestions and nitpicks, but overall a nice implementation.

Comments formatted as [conventional comments](https://conventionalcomments.org/).

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Great question. I'll fix.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Agreed.  Will fix.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Not sure.  But deserializeList seems more semantically reasonable, so I'll switch to that.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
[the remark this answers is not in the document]
mgaffigan: @kpalang, not sure what you are referring to by that.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Agreed.  Fixed.

server/src/com/mirth/connect/model/filters/elements/ContentSearchElement.java
mgaffigan: Ooh!  I didn't realize that was available in Java 17.  Absolutely.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
mgaffigan: Ah... fine.  I guess I'll eat my broccoli.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
mgaffigan: Copy/paste error.  Fixed.

server/src/com/mirth/connect/model/filters/MessageFilter.java
mgaffigan: Out-of-date habits.

server/src/com/mirth/connect/model/filters/MessageFilter.java
mgaffigan: It's what was [checked in originally](https://github.com/mgaffigan/oie/blob/main/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java#L786-L790).  I agree it should be fixed, though.  Not critical, but I'm here.

server/src/com/mirth/connect/model/filters/MessageFilter.java
mgaffigan: It's what was [checked in originally](https://github.com/mgaffigan/oie/blob/main/client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowser.java#L786-L790).  Moving it to separate append calls.

server/src/com/mirth/connect/model/filters/MessageFilter.java
mgaffigan: I prefer a linter be used to settle whitespace wars, but I've added a linebreak.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
jonbartels: #321 

#322

jonbartels (approved):

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
```diff
@@ -510,6 +514,81 @@ public Boolean hasAdvancedCriteria() {
...
+        }
+    }
+
+    private String formatMetaDataValue(Object value) {
+        if (value == null) {
+            return "";
+        }
+
+        if (value instanceof Calendar calendar) {
+            return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(calendar.getTime());
```
kpalang, line 586: thought (non-blocking): I found this cool site with modern approaches to things in Java, and one of them is to use [`DateTimeFormatter` instead of `SimpleDateFormat`](https://javaevolved.github.io/datetime/date-formatting.html).

kpalang (approved):
Approved.

I found one small improvement after the previous review, but it is small and I'm not gonna request a change for this.

NicoPiel (approved):
Great stuff.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserRecentFilterStore.java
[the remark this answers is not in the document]
mgaffigan: Resolving given kpalangs approval on 2026-06-30.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
mgaffigan: Noted.  This is copy/paste from prior location, but I do like it for future.

client/src/com/mirth/connect/client/ui/browsers/message/MessageBrowserAdvancedFilter.java
mgaffigan: Resolving given kpalang approval 2026-06-30.

tonygermano (commented):
This non-blocking, but the ToString, Equals, and Hash builders in lang3 could probably reduce the lines of code and avoid manual string building.

I approve of the recent filters being saved in memory only. If we persist them in the future, I think it should be a database-backed user preference rather than a local java preference.

Either way it could get complicated if the user has more than one admin client open at the same time.

A future feature request could be the ability to pin or "forget" a recent search to have a little more control in keeping useful requests around longer.

PR #280, opened 2026-03-30
Add certificate pinning to address MITM vulnerabilities

mgaffigan:
Addresses MITM vulnerability by adding `-trust {config}` option to client.  Adds `administrator.pinnedclienttrust` option to mirth.properties to control the option.

| Option | Description | Hostname Validation |
| :--- | :--- | :--- |
| `pki` | Trust CAs and peer trust based on client JVM config. Normal trust mode. | Required (matches CN or SAN) |
| `webserver` | Trust the auto-generated server's certificate. | Not Validated |
| `<thumbprint>` | Trust a specified SHA-256 certificate thumbprint. | Not Validated |
| `insecure_trust_all_certs` | Trust all certificates without validation (not recommended). | Not Validated |
| `localhost` | Trust all certificates for the hostname localhost | Not Validated |

If argument is not passed to the client (e.g.: for development without a launcher), `pki,localhost` is used.  If the option is not specified in mirth.properties or for new installs, `pki,webserver` is used.  Multiple options can be specified by separating with commas.

Typical scenarios:

- Localhost, docker, or direct network access with trusted or self signed certificate<br>✅ No edits needed, secure by default
- Reverse proxy with trusted certificate<br>✅ No edits needed, secure by default
- Reverse proxy with self signed certificate<br>⚠️ Edit needed to add the proxy's certificate or enable `insecure_trust_all_certs`

### DIFF
--- a/client/src/com/mirth/connect/client/ui/CommandLineOptions.java
+++ b/client/src/com/mirth/connect/client/ui/CommandLineOptions.java
@@ -3,6 +3,10 @@
 
 package com.mirth.connect.client.ui;
 
+import java.util.ArrayDeque;
+import java.util.Arrays;
+import java.util.Deque;
+
 import org.apache.commons.lang3.StringUtils;
 
 /**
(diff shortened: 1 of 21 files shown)


ChristopherSchultz:
I see only one useful setting: "use this specific trust store" with a proper certificate in it, having been fetched and approved by the launcher before the client runs.

"Trust any localhost" isn't worth implementing. "Trust anybody" isn't worth implementing. Thumbprints require more code to implement your own trust store. Just use the existing SSLSocketFactory tie-ins that allow you to specify the trust managers. Writing your own trust manager that looks at the fingerprint is very non-Java.

I would drop all the server-specific parts of this PR and all the client stuff with the exception of anything that points it at the correct trust store.

mgaffigan:
@ChristopherSchultz 
> "Trust any localhost" isn't worth implementing. 

Trust any localhost is intended to be used without a launcher.  localhost is by nature not vulnerable to MITM attack, nor loss of confidentiality.  This option primarily exists for developers, and is not present in the default configuration.

> "Trust anybody" isn't worth implementing. 

Trust anybody is intended for use in allowing backwards compatibility with the historical behavior.  It is fundamentally insecure if someone modifies their server configuration to use that option, but I believe it is neccessary for practical purposes in a brown-field codebase.  That's also why it has the name `insecure_trust_all_certs` with the word "insecure" in the name.

> Thumbprints require more code to implement your own trust store. Just use the existing SSLSocketFactory tie-ins that allow you to specify the trust managers. Writing your own trust manager that looks at the fingerprint is very non-Java.

The "pki" trust option is the "java defaults" trust store.  For anyone with the ability to execute in that mode, they are welcome to.  There is no mandate to use the trust managers you disagree with.

> I would drop all the server-specific parts of this PR and all the client stuff with the exception of anything that points it at the correct trust store.

How do you imagine the launcher should know what servers to trust without a list provided in the JNLP?  Are you assuming that the JNLP is served from the singular valid cert?  Or some OOB configuration

I'm very open to other designs, but this is the only one I see that is backwards compatible with current launchers, while moving us to better security.  If you have an alternate model, send PR's or document a design compatible with the ecosystem of launchers.

mgaffigan:
Related https://github.com/kayyagari/ballista/pull/53

pacmano1:
@mgaffigan at least with Launcher which prompts about accepting the server cert (but no longer verfies jars),  is this now addressed?

mgaffigan:
> @mgaffigan at least with Launcher which prompts about accepting the server cert (but no longer verfies jars),  is this now addressed?

No. This is still required to address the issue. Both launcher and client are vulnerable.

PR #282, opened 2026-03-31
cache ResultSet metadata in getMetaDataMapByMessageId to reduce per-row metadata lookups

NicoPiel:
Cache ResultSetMetaData (column count, types, and names) before iterating rows so getColumnType/getColumnName are not called for every row. Adjust column indexing and decryption logging to use the cached metadata. This reduces repeated metadata calls and improves performance when reading connector metadata from the database.

### DIFF
--- a/donkey/src/main/java/com/mirth/connect/donkey/server/data/jdbc/JdbcDao.java
+++ b/donkey/src/main/java/com/mirth/connect/donkey/server/data/jdbc/JdbcDao.java
@@ -2935,8 +2935,6 @@ public class JdbcDao implements DonkeyDao {
             statement = connection.prepareStatement(querySource.getQuery("getMetaDataMapByMessageId", values));
             resultSet = statement.executeQuery();
 
-            // Cache ResultSet metadata before the loop — the schema is the same for every row,
-            // so calling getMetaData()/getColumnType()/getColumnName() per row is wasteful.
             MetaDataColumnType[] columnTypes = null;
             String[] columnNames = null;
             int columnCount = 0;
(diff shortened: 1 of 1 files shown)


donkey/src/main/java/com/mirth/connect/donkey/server/data/jdbc/JdbcDao.java
```diff
@@ -2935,7 +2935,22 @@ private Map<Long, Map<Integer, Map<String, Object>>> getMetaDataMaps(String chan
...
+                    ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
+                    columnCount = resultSetMetaData.getColumnCount();
+                    columnTypes = new MetaDataColumnType[columnCount];
+                    columnNames = new String[columnCount];
+                    for (int i = 0; i < columnCount; i++) {
+                        columnTypes[i] = MetaDataColumnType.fromSqlType(resultSetMetaData.getColumnType(i + 1));
+                        columnNames[i] = resultSetMetaData.getColumnName(i + 1).toUpperCase();
+                    }
+                }
+
```
mgaffigan: Eager initialize columnTypes and columnNames to avoid null and null check.  You can call `getMetaData` before `next`.

You don't need to cache things like `getColumnCount()`.  Any driver that's making a round trip for that is fundamentally broken.

mgaffigan (changes requested):
I doubt this is making actual round-trips, but unopposed to cleaning things up.  At a minimum, we should drop some GC pressure from the repeated `toUpperCase` and `fromSqlType`.

Also: consider whether the move to 0-indexing is actually adding value.  For the cost of two empty array elements, this PR gets a lot smaller and there's less opportunity to screw up the zero-based vs one-based indexing.

Also: missing DCO

donkey/src/main/java/com/mirth/connect/donkey/server/data/jdbc/JdbcDao.java
NicoPiel: Fixed

mgaffigan (approved):

jonbartels (approved):
I also did some digging to find out if this is a DB hit or something else. It reduces 1000s of calls to only one call per column. Its a good change. Great find Nico!

kayyagari (approved):
Thank you for spotting it.

kryskool (approved):

PR #286, opened 2026-04-05
Avoid having the version spread everywhere

mgaffigan:
Centralizes to the existing `mirth-build.properties` `version` property.

Shifts tests to using 99.99.99 as the version to avoid confusion.

### DIFF
--- a/server/src/com/mirth/connect/connectors/dimse/destination.xml
+++ b/server/src/com/mirth/connect/connectors/dimse/destination.xml
@@ -2,7 +2,7 @@
 	<name>DICOM Sender</name>
 	<author>NextGen Healthcare</author>
 	<pluginVersion>@mirthversion</pluginVersion>
-	<mirthVersion>4.5.2</mirthVersion>
+	<mirthVersion>@mirthversion</mirthVersion>
 	<url>https://openintegrationengine.org</url>
 	<description>This connector allows Open Integration Engine to send messages using DICOM protocol.</description>
 	<clientClassName>com.mirth.connect.connectors.dimse.DICOMSender</clientClassName>
--- a/server/test/resources/notifications-populated.json
+++ b/server/test/resources/notifications-populated.json
@@ -3,7 +3,7 @@
     "url": "https://api.github.com/repos/nextgenhealthcare/connect/releases/176445080",
     "assets_url": "https://api.github.com/repos/nextgenhealthcare/connect/releases/176445080/assets",
     "upload_url": "https://uploads.github.com/repos/nextgenhealthcare/connect/releases/176445080/assets{?name,label}",
-    "html_url": "https://github.com/nextgenhealthcare/connect/releases/tag/4.5.2",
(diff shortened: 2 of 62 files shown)


server/test/resources/notifications-populated.json
```diff
@@ -27,19 +27,19 @@
...
+    "tarball_url": "https://api.github.com/repos/nextgenhealthcare/connect/tarball/99.99.99",
+    "zipball_url": "https://api.github.com/repos/nextgenhealthcare/connect/zipball/99.99.99",
+    "body_html": "<p>Mirth® Connect 99.99.99 is a patch release that includes enhancements, bug fixes, and security improvements. For more details, visit our See What's New and Release Notes pages linked below.</p>\n<p><strong>NOTICE: In 2025, Mirth® Connect will move to Java 17 as the minimum supported Java version.</strong></p>\n<p><a href=\"https://github.com/nextgenhealthcare/connect/wiki/99.99.99---What's-New\">See What's New</a> | <a href=\"https://github.com/nextgenhealthcare/connect/wiki/Upgrading\">Upgrade Guide</a> | <a href=\"https://github.com/nextgenhealthcare/connect/milestone/85?closed=1\">Release Notes</a></p>",
```
jonbartels, line 42: Nitpick - If we are removing specific versions should we still have a link to a specific milestone in this file?

server/src/com/mirth/connect/connectors/dimse/destination.xml
```diff
@@ -2,7 +2,7 @@
 	<name>DICOM Sender</name>
 	<author>NextGen Healthcare</author>
 	<pluginVersion>@mirthversion</pluginVersion>
-	<mirthVersion>4.5.2</mirthVersion>
+	<mirthVersion>@mirthversion</mirthVersion>
```
jonbartels, line 5: Is `@mirthversion` just a static string or is it substituted somewhere at compile time?

jonbartels (changes requested):

server/src/com/mirth/connect/connectors/dimse/destination.xml
mgaffigan: It's substituted at build time. It was present already for pluginVersion, but was not in use for mirthVersion. Not sure why.

server/test/resources/notifications-populated.json
mgaffigan: It's just test data for a unit test. Given infinite resources, we'd strip it to entirely false data, but it is not used anywhere. Only point to changing it is to avoid "real" version numbers showing up anywhere they are not in actual use.

jonbartels (approved):

jonbartels (approved):

kryskool (approved):
Ok, it works for me

tonygermano (approved):
Looks good to me. For anyone else searching like I did, this is where the replace happens:

https://github.com/OpenIntegrationEngine/engine/blob/5203b3a38b611ed21c5513ae337e450bd5a58410/server/build.xml#L444-L447

NicoPiel (approved):

Issue #197, opened 2025-10-28
[IDEA] Client launches slow


mgaffigan:
I feel like the client might be slow to launch because it makes 50x requests to the API below on launch once for each plugin.  If this could be made to be a single round-trip, that would probably speed things up.  No one likes waiting on launch.

```http
GET http://127.0.0.1:8080/api/4.5.2/extensions/Channel%20Reader/enabled HTTP/1.1
X-Requested-With: openintegrationengine-client
Accept: application/xml
Accept: application/json
User-Agent: Jersey/2.22.1 (Mirth Server Connection)
Host: 127.0.0.1:8080
Connection: Keep-Alive
Cookie: JSESSIONID=node05i755oi7lwba1whm5458j7rfn0.node0
Accept-Encoding: gzip,deflate


HTTP/1.1 200 OK
Date: Sat, 18 Oct 2025 15:51:00 GMT
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: false
Access-Control-Allow-Methods: GET, POST, DELETE, PUT
Access-Control-Allow-Headers: Content-Type
Content-Security-Policy: frame-ancestors 'none'
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
Strict-Transport-Security: max-age=31536000; includeSubDomains
Content-Type: application/xml
Content-Length: 23

<boolean>true</boolean>
```

PR #287, opened 2026-04-05
Speed up jnlp launch by 58%

mgaffigan:
Speeds launch time by not requiring multiple round-trips for each plugin.  Classpath with ballista remained identical.  No obvious benefit to the historical design - presumed to be some Java Security Manager pipe dream.

| Scenario | Before | After | % change |
|---|---|---|---|
| Cold cache, 100ms latency | 8.9s | 3.7s | ↓58% |
| Warm cache, 100ms latency | 5.8s | 0.7s | ↓87% |

Mildly related: #197


pacmano1 (approved):

NicoPiel (approved):

gibson9583 (approved):

tonygermano (commented):
Looks fine at a glance. I think the only thing we're really losing is the title and vendor information for the individual extensions.

Does this still allow launching from MCAL?

mgaffigan:
> Looks fine at a glance. I think the only thing we're really losing is the title and vendor information for the individual extensions.

The JNLP data is not shown anywhere that I can find in any of the launchers I've tested.  The vendor information still shows normally in Extensions.

> Does this still allow launching from MCAL?

It works fine, but MCAL is broken in main - I recommend we drop support entirely.

PR #288, opened 2026-04-06
Feature/issue 261 export tags  -Prune unrelated channel IDs from exported channel tags

eddcomeau:
## Summary
  Prunes exported channel tag membership so channel exports only include the exported channel's ID in each tag.

  ## Changes
  - prune tag channel IDs in server-side export assembly
  - prune tag channel IDs in client-side channel export
  - update ChannelServletTest expectations for exported tag membership

  ## Verification
  - ChannelServletTest passes when run directly (It was edited as part of this commit)
  - Rebuilt with Ant -DdisableTests=true
  - Tested the UI (windows 10) by exporting a single channel, importing, and exporting a group, importing.

### DIFF
--- a/client/src/com/mirth/connect/client/ui/ChannelPanel.java
+++ b/client/src/com/mirth/connect/client/ui/ChannelPanel.java
@@ -1928,7 +1928,9 @@ public class ChannelPanel extends AbstractFramePanel {
         List<ChannelTag> channelTags = new ArrayList<ChannelTag>();
         for (ChannelTag channelTag : getCachedChannelTags()) {
             if (channelTag.getChannelIds().contains(channel.getId())) {
-                channelTags.add(channelTag);
+                ChannelTag exportTag = new ChannelTag(channelTag);
+                exportTag.setChannelIds(Collections.singleton(channel.getId()));
+                channelTags.add(exportTag);
             }
         }
 
--- a/server/src/com/mirth/connect/server/api/servlets/ChannelServlet.java
+++ b/server/src/com/mirth/connect/server/api/servlets/ChannelServlet.java
@@ -312,6 +312,11 @@ public class ChannelServlet extends MirthServlet implements ChannelServletInterf
             channel.getExportData().setChannelTags(channelTags
                     .stream()
                     .filter(tag -> tag.getChannelIds().contains(channel.getId()))
+                    .map(tag -> {
+                        ChannelTag exportTag = new ChannelTag(tag);
+                        exportTag.setChannelIds(Collections.singleton(channel.getId()));
+                        return exportTag;
+                    })
                     .collect(Collectors.toList()));
 
             channel.getExportData().setDependencyIds(channelDependencies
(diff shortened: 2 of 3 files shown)


mgaffigan (dismissed):

kryskool (approved):

NicoPiel (dismissed):

client/src/com/mirth/connect/client/ui/ChannelPanel.java
```diff
@@ -1924,17 +1924,19 @@ private void addDependenciesToChannel(Channel channel) {
...
+            if (channelTag.getChannelIds().contains(channel.getId())) {
+                ChannelTag exportTag = new ChannelTag(channelTag);
+                exportTag.setChannelIds(Collections.singleton(channel.getId()));
+                channelTags.add(exportTag);
+            }
+        }
+
+        if (CollectionUtils.isNotEmpty(channelTags)) {
+            channel.getExportData().setChannelTags(channelTags);
+        }
```
tonygermano: chore: there appears to be a whitespace issue here

Most of this method did not change.

tonygermano (changes requested):
I have a few questions about these changes and the associated issue.

1. Where is the import code that looks at this information? Presumably it would need to create the tag in some situations and not create it in others. Can we be certain that this will not cause unwanted behavior in some scenarios?
2. The issue suggests emptying the channel list from the exported tag entirely. Is there a reason to keep the current channel Id rather than using an empty Set?
3. The issue also identifies a problem with the tags reporting incorrect channel values/counts. Can we confirm why that was the case, and whether this PR resolves that issue or not?

eddcomeau:
> I have a few questions about these changes and the associated issue.
> 
> 1. Where is the import code that looks at this information? Presumably it would need to create the tag in some situations and not create it in others. Can we be certain that this will not cause unwanted behavior in some scenarios?
> 2. The issue suggests emptying the channel list from the exported tag entirely. Is there a reason to keep the current channel Id rather than using an empty Set?
> 3. The issue also identifies a problem with the tags reporting incorrect channel values/counts. Can we confirm why that was the case, and whether this PR resolves that issue or not?

Thanks 

1) 
[The serialized list of channels is really only used for NEW tags.](https://github.com/OpenIntegrationEngine/engine/blob/be058f69ff8f181c8f4ac0d39db955cb265853dc/client/src/com/mirth/connect/client/ui/ChannelPanel.java#L660-L687)
If it's an existing tag it could only ever add the imported channel. 
This PR would change that behavior a bit, so brand new tags used to get added to every channel in the serialized list, now a brand new tag will only get added to the single channel you are importing. I think the only case you get this is if you export a channel with a tag, delete the tag everywhere, then re-import the channel it will now only add the tag to the imported channel than adding across all channels. So, I think this change would avoid unwanted side effects. 

2) Using an empty set would work for existing tags but would likely cause an error for new tags. I could change the import, but I think if I export a channel with a tag, I expect that tag to be there when I import it, even if it means making a new one. 

3) To be honest I couldn't replicate that. Although I believe because NEW tags were being added to the entire serialized list of channels, non-existent channels could still be added to the list. Throwing off the number. My hope was that the commit is a good move and if it resolved this that would be a bonus. But let me know if you disagree and I'm willing to dig more.

client/src/com/mirth/connect/client/ui/ChannelPanel.java
eddcomeau: I fixed this up in the newer one

eddcomeau:
Rebased/updated against current `main` and rechecked the export-tag behavior.

I believe the current approach is still preferable to exporting an empty channel ID set because it preserves the expected “imported channel keeps its tag” behavior for newly-created tags, while preventing unrelated channel IDs from being serialized into the exported channel.

@tonygermano , does that address the concern well enough to clear the change request, or would you prefer I adjust the import path too?

PR #299, opened 2026-04-16
feat: add optional dcm4che5 DICOM backend with runtime selection5 backend

jessesaga:
I realize this is a large changeset but I'm not sure how this can be done incrementally. I considered splitting it, but the abstraction layer is dead code without a second backend, and the variant-filtered JAR loading only exists to pick between the two. Open to restructuring if there's something that would make review easier.

Why: at Saga IT we've had to work around limitations in the dcm4che2 implementation and we believe the newer dcm4che5 version will have fewer bugs and potentially allow room for performance improvements - not to mention receiving security updates.

Caveat: only tested locally end-to-end, including mTLS DIMSE channels on both backends. This code has not been tested in real world workloads.

## Summary

Adds an optional dcm4che5 backend to the DICOM Listener / Sender connector alongside the existing dcm4che2 backend. Selection is per-server via a new `dicom.library` property in `mirth.properties`. The default remains `dcm4che2` - existing installations see zero behavioral changes on upgrade.

dcm4che2 has not had a release since 2015. dcm4che5 is actively maintained, has better support for modern Java and TLS, and receives security fixes. This PR gives users the option to move forward without forcing the migration.

## Backward compatibility

The default codepath is entirely unchanged. For installations that don't opt in:

- `dicom.library`, unset, defaults to `dcm4che2`
- Channel XML unchanged - no field renames, no new required properties
- Existing custom `DICOMConfiguration` classes continue to load (with fallback logic for pre-PR implementations)
- dcm4che2 library JARs still load and behave identically
- One minor API change on `DICOMUtil`: `byteArrayToDicomObject()` now returns `OieDicomObject` instead of `DicomObject`. Duck-typed scripts work unchanged thanks to Object-VR overloads. Scripts doing explicit `(DicomObject)` casts or `instanceof DicomObject` checks need `.unwrap()` - documented in the migration guide.

## Approach

- **Version-neutral abstraction** in `com.mirth.connect.connectors.dimse.dicom`: `OieDicomObject`, `OieDicomElement`, `OieDicomSender`, `OieDicomReceiver`, `OieDicomConverter`, `OieVR`. Decouples Mirth's DICOM code from a specific library version.
- **Two implementations**: `dcm2/` wraps existing `MirthDcmSnd`/`MirthDcmRcv` (unchanged in behavior). `dcm5/` composes from `Device` + `Connection` + `ApplicationEntity` + service handlers in idiomatic dcm4che5 style.
- **Runtime selection**: `DicomLibraryFactory` reads `dicom.library` at startup and instantiates the matching backend via reflection. Default is `dcm4che2`.
- **Variant-filtered JAR loading**: Extension library entries gain an optional `variant="dicom.library:<value>"` attribute. `MirthLauncher` (server) and `WebStartServlet` (Administrator) both honor it, so only the active backend's JARs are loaded on the server and shipped to the Administrator.

## Testing

- **Unit**: 768 tests across the repo, 0 failures. Parity suites for converter, object, element, receiver, sender on both backends.
- **Integration**: C-STORE, C-ECHO, storage commitment, TLS, and cross-library round-trip tests under `server/test/.../dimse/dicom/integration/`.
- **Manual**: Two-channel DIMSE forwarding with mutual TLS, verified on both backends via the official Mirth Administrator Launcher.

## Migration

Migration guide for users opting into dcm4che5: [`docs/dcm4che5-migration-guide.md`](docs/dcm4che5-migration-guide.md). Covers enable/disable, element-name difference (keyword vs PS3.6), async C-STORE dispatch, TLS cipher configuration, and three UI settings that are no-ops on dcm5 (two `bufSize` tuning flags; `dest`, which is in fact a silent no-op on dcm4che2 upstream as well).

## Dependencies

Adds `dcm4che-core-5.34.3` and `dcm4che-net-5.34.3` (MPL-1.1). License entries added to `server/docs/thirdparty/THIRD-PARTY-README.txt`.

## Not in this PR

- Does not remove dcm4che2 - default unchanged.
- Does not add dcm4che5-specific features beyond the existing DICOM feature set.

## Manual testing approaches taken beyond test cases

- Fresh upgrade on an existing install (no `dicom.library` set): existing DIMSE channels deploy and process messages identically to pre-upgrade
- Set `dicom.library = dcm4che5`, restart: channels re-deploy cleanly, C-STORE and C-ECHO behavior verified
- Toggle back to `dicom.library = dcm4che2`, restart: behavior returns to default
- Mutual-TLS DIMSE between two channels on each backend
- Administrator Launcher connects cleanly and loads only the active backend's JARs


jonbartels (changes requested):
Should the copyright headers be updated? If the code comes from Mirth, then they are still Mirths copyright. If they are new files generated by the contributors, then having the contributor copyright would be appropriate.

The headers should also be updated to the `SPDX` format seen elsewhere in the project (bonus points for adding a linter check for this :D )

What real world tests would be most valuable for testing this code?

jessesaga:
Good point on the Copyright and SPDX format. I'll update the newly added files with the SPDX format and the Saga IT, LLC copyright.

What's the policy for the existing files? Convert to SPDX and add an additional FileCopyrightText line?

Not sure about the linter but I can take a look :)

In terms of real world tests, I'd say the biggest current gaps are around load testing and interop with other PACS. 

I've done small local load tests but nothing like the volume a real world system might see. It's possible that sustained throughput could reveal some issues. (especially with different threading and queueing settings)

I have not yet tested against any real PACS - just Mirth to Mirth communication. It would be good to do some testing against at least Orthanc and dcm4chee-arc.

mgaffigan:
I'll be the first to call out my ignorance regarding dicom, but is there a reason this is all in the same plugin?

Seems like only one version can run at a time, and there's a bunch of proxies to select between the two at runtime (vs loading the legacy dcm4che2 plugin, or loading the alternate dcm4che5 plugin).  That presumably introduces a fair bit of debt for go-forward, and change risk for back-compat.

jessesaga:
The main reason is to provide an easy upgrade path from the ancient unmaintained dcm4che2 libraries and this seemed like a good way to do it. I'm open to other approaches but this way makes the upgrade relatively painless, the abstraction layer provides a little defense against future dcm4che api changes, and the mirth.properties and runtime library resolution keeps only one of the implementations on the server classpath and client classpath at a time.

mgaffigan:
Sure.  I think the salient questions would be: 

1. Are any changes in behavior intended for dch4che2?
2. Can this PR be substantially simplified by avoiding the proxies and runtime selection?
3. Is it required/realistic to expect new development to operate compatible to dcm4che2/dcm4che5 (with the new wrappers)?  Or are the differences practically sufficient to make any code coupled to one version or the other.

The user effect seems neutral: mirth.properties change vs plugin change.

mgaffigan:
From some hacking around by stripping out the wrappers and dcm4che2 and making v2 vs v5 a separate plugin:

* PR to extract DICOM from core server build (remove dep on dcm4che2 et al)<br>+126/-47
* PR to add separate dcm4che5 plugin without wrappers<br>+1,800/-336

Vs this PR which is at +4,616/-399 for implementation.

I call out again: I don't have the DICOM background to evaluate how helpful the wrappers are.  My own experience leans towards avoiding wrappers unless they add significant value (since it makes it "OIE DICOM" instead of standard types) - but I can't estimate the value here.

jessesaga:
@mgaffigan - The primary goal with the current approach is to ensure that it is perfectly backward compatible to make sure it's an uneventful upgrade process for users.

> Are any changes in behavior intended for dch4che2?

No, a lot of care was taken to ensure that users that not only just receive and send DICOM can simply upgrade and have things work with dcm4che5. The migration guide in the PR outlines a couple of minor API changes. Most users will be able to switch to dcm4che5 from 2 and have it work without changes. If they are extensively using the mirth DICOM API then it's possible they might have to make very minor adjustments.

To find the relevant section in the guide you can search for this phrase - "Only these specific patterns require a one-line change"

> Can this PR be substantially simplified by avoiding the proxies and runtime selection?

Yes, some, but we'd lose Mirth's excellent backward compatibility story. If we just ripped out dcm4che2 and plugged in dcm4che5 the most minimal changes possible many DICOM listener/destination users would be forced to rework their code.

> Is it required/realistic to expect new development to operate compatible to dcm4che2/dcm4che5 (with the new wrappers)? Or are the differences practically sufficient to make any code coupled to one version or the other.

There are very minimal changes that may be required depending on how extensively the DCIOM API is being excercised. The main difference is around java types that are passed but each type is actually duck-type compatible and the javascript engine will be able to resolve correctly. Any explicit casts would have to be updated if they're being used.

mgaffigan:
That all still seems compatible with having alternate v2 and v5 plugins (with only one installed at a time).  I'd suggest this be done that way to avoid reinventing the wheel and making yet another dicom-specific modification to the main server.

Using actual plugins has the benefit of:
1. Making the risk to any dcm4che2 users nearly zero
2. Avoiding new complications to MirthLauncher and WebStartServlet
3. Avoiding the runtime indirection between v2 and v5 (instead using standard plugin loading)
4. Using standard plugin add/remove processes to pick between v2 and v5

I defer to your judgement that the wrappers themselves are required to maintain backwards compatibility with user scripts.

I'm happy to talk through that in more detail if you would like.  Rough process would be:

1. Pull DICOM build out of the main server build (I'd propose `plugins/dicom/build.xml` as a new home).
    - Move the source and libs out of server/
    - Add a separate javac call and move the jar calls
    - [Fix the inverted dependency](https://github.com/mgaffigan/oie/commit/6f36620fe25a97e038519d0994ca2ac906aea6df) the server has on the DICOM plugin
2. Add the new wrappers and code from this PR as a new plugin (I'd propose `plugins/dicom5/` or similar).  The dcm4che2 indirection should be omitted, since those are handled in the separate legacy plugin codebase.  Only the implementation and dcm4che5 backcompat shims would be required.
3. Update installer to update the version found (v2 for v2, v5 for new installs or if v5 is found)

PR #316, opened 2026-05-27
Eliminate O(n^2) status removal in DashboardTreeTableModel

NicoPiel:
updateChannelNodes called statuses.remove(status) inside a loop over
child nodes. With an ArrayList, each remove is O(n), making the update
O(channels × statuses). Collected matched statuses in a HashSet and
called removeAll once after the loop, reducing to O(n + m).

### DIFF
--- a/client/src/com/mirth/connect/client/ui/DashboardTreeTableModel.java
+++ b/client/src/com/mirth/connect/client/ui/DashboardTreeTableModel.java
@@ -498,13 +498,14 @@ public class DashboardTreeTableModel extends SortableTreeTableModel {
     }
 
     private void updateChannelNodes(Map<String, DashboardStatus> statusMap, List<DashboardStatus> statuses, MutableTreeTableNode parent) {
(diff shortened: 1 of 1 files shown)


jonbartels:
Is there an existing unit test that covers this method? 

If no, can you create one?

mgaffigan (approved):
This looks fine, but was this showing up on a profiler?  There's going to be a _lot_ of changes if you are not chasing based on actual performance issues.

kpalang (approved):

kryskool (approved):
Good catch, works for me

tonygermano (commented):
Similar to Mitch's comment, is there a measurable improvement from this change? The old code wasn't wrong. I don't want to turn down improvements, but I also don't want to encourage optimizations with no measurable impact, as they take up PR review bandwidth that could be used elsewhere.

NicoPiel:
I haven't measured anything; probably only measurable with a huge dashboard. I was looking for potential optimisations to the dashboard and stumbled upon this.

PR #317, opened 2026-05-27
Move getMaxMessageId off EDT in MessageBrowser.runSearch

NicoPiel:
getMaxMessageId is a blocking HTTP call that was executed on the Swing
EDT, freezing the UI while waiting for the server response. Moved it
into SwingWorker.doInBackground(); all subsequent UI updates run in
done() back on the EDT.


mgaffigan (changes requested):
Need protection against reentrance, otherwise double-click will race with indeterminate state.

PR #318, opened 2026-05-27
Cache compiled regex Pattern in MirthTree

NicoPiel:
Pattern.compile(" (\\(.*\\))") was called on every iteration of two
tree-traversal while-loops in constructVariableName and
constructNodeDescription. Pattern compilation is expensive and the
pattern never changes. Extracted to a static final field.

Signed-off-by: Nico Piel <nico.piel@hotmail.de>


jonbartels (approved):

mgaffigan (approved):

kryskool (approved):
Works for me

tonygermano (approved):

PR #319, opened 2026-05-27
Build reverse index for tag-channel lookup in TagTreeCellRenderer

NicoPiel:
Each cell render previously iterated all cached tags O(n) to find tags
for a channelId. With many tags and frequent repaints (scroll, refresh),
this made the dashboard sluggish O(channels x tags) per paint cycle.

Now the renderer builds a Map<channelId, List<ChannelTag>> on first use
and whenever the tag set reference changes (i.e. after each tags refresh).
Render-time tag lookup is O(1); sort happens once at index-build time.


mgaffigan (changes requested):
Please confirm invalidation path.  Is this really showing up on profiling?  Seems like all of this should be in memory and negligibly fast.

PR #54, opened 2025-04-23
Migration to Gradle

NicoPiel:
I've started working on migrating OIE from Ant to Gradle.

To test, run `./gradlew clean build -x test`

(tests are failing)

Fixes #052

PR #326, opened 2026-06-11
Replace the Ant build with Gradle, output verified byte-identical

pacmano1:
Refs #52.

This replaces the Ant build with a standard Gradle build in one verified step. The product the build produces did not change, and that claim is checkable: the output of both builds was compared archive by archive, file by file. This description is the full write-up (it used to live in `BUILD-MIGRATION.md` in the repo; per review it now lives here, since it describes this change rather than the ongoing build).

## What changed since the first review

Addressing the review feedback (thanks @mgaffigan):

- **The verification tooling moved out of this repo** to [pacmano1/oie-build-parity](https://github.com/pacmano1/oie-build-parity) (`compare_builds.py`, the jar-provenance audit, `snapshot_distribution.py`). It is migration scaffolding, not part of the build, so it no longer adds to this PR's review surface. **If the maintainers would rather it live under the org, say the word and I'll move it to `OpenIntegrationEngine/oie-build-parity`.** It's offered, not imposed.
- **`BUILD-MIGRATION.md` is removed**; this write-up is now the PR description (it would only go stale sitting in the tree).
- **Build flags are now Gradle project properties (`-P`), and `-D` support is gone.** A clean breaking change, as suggested, rather than carrying a compatibility shim. `-PdisableSigning=true`, `-Pcoverage=true`, `-Pversion=x.y.z`, and `-Pkey.storepass=<PIN>` for signing. The Dockerfile and CI were updated to match.
- **The hand-written empty `package-info.class` generation is gone.** Ant's javac emitted empty, inert marker classes for the three annotation-free `package-info.java` sources; the Gradle build no longer reproduces them. This is now a documented, zero-runtime-effect difference (the comparison tool classifies it), instead of being faked with a hand-built class file.
- **A few incidentals**: the in-build `snapshotDistribution` task moved to the external tooling; the `writeArtifactInventory` task description was clarified; some unrelated personal entries were removed from `.gitignore`.

The substance of the migration below is unchanged.

## The short version

- The build system changed from Ant to Gradle. **The product it builds did not change**, and that claim is checkable.
- 374 of the 419 library jars that used to be stored in the repository are now downloaded from Maven Central, each accepted only after proving it is **the exact same file, bit for bit** (same SHA-1). The other 45 stay in the repository, each with a documented reason (appendix).
- Every check on the build's output is reproducible, so any reviewer can re-run the verification rather than trust this description. The recipe is at the bottom; the tooling is in [oie-build-parity](https://github.com/pacmano1/oie-build-parity).
- Three commits, reviewable separately: 1,806 pure file moves to the standard layout (`git show --stat -M100%` confirms zero content change), the build swap, and the dependency swap (plus a Windows-docs note from @jonbartels on top). Only the head commit builds; squash-on-merge is fine.
- Substantial parts of this work were produced with AI assistance as a process experiment. The controls that made that safe, and the errors those controls caught, are described below.

## Why one change instead of piecemeal

None of the direction here is new. @NicoPiel proposed the move to a dependency-managed build (#52) and drove three successive Gradle attempts (#54, #100, #214); @jessesaga built a fourth (#243). In those reviews @mgaffigan named the bar a one-step replacement would have to clear, byte-identical output, and wrote the verification toolkit this branch is cross-checked with. The three-commit structure follows the reviewable-commit standard Tony Germano held the project's last large dependency change (#146) to. This branch is built on that groundwork.

The incremental path (wrap the old build first, convert one piece at a time) is the default for a reason: replacing a build system in one step is reckless **when you can't prove the output is the same**.

This branch takes the other path because the risk became measurable. The bar is met, and here is exactly what "identical" means, including everything that is *not* identical:

- **Every archive in the distribution has identical contents** to an Ant build of this branch's parent commit (`8c1111ba3`), apart from build-tool fingerprints rather than product content: (1) 121 archives carry an Ant-version label Gradle doesn't write; (2) the build-written file timestamps inside those same archives differ, because both tools stamp a fixed fake date and chose different ones (third-party jars pass through byte-identical); (3) one small file (`version.properties`) embeds the build date, so it differs between any two builds; (4) two jars contain one extra folder *entry* (not a file); and (5) the **empty `package-info.class` markers** Ant's javac emitted for the three annotation-free `package-info.java` sources are not reproduced (inert, no runtime effect; see "What changed since the first review"). Every class file, resource, script, and configuration byte is the same. The comparison tool classifies each of these categories itself; nothing was waved through by hand.
- **All 654 tests pass, and the list of test classes that ran is identical** to the Ant build's. (The CI test counter drops to roughly half its historical number because the old build accidentally wrote a duplicate results file that counted every test twice; the true count was always 654.)
- **The published v4.6.0-rc1 release was cross-checked** against a local Ant build of the same source: 473 of 490 archives bit-for-bit identical in content. The 17 that differ are all the release pipeline, not compilation (the signing step repackages modular jars, rewriting one bookkeeping file; signature files and installer files can never match between builds). None differ in compiled code.
- **The signing path works end to end**: all 216 jars that ship signed are signed by the new build and pass `jarsigner -verify` (strict mode adds the expected self-signed-dev-keystore warning; the old build warns identically).
- **The built product boots and runs** (a human boot test of the Gradle-built `server/setup`; a sanity check on top of the archive comparison).

A verified cutover is one reviewable event with the proof attached, still split into three commits so each concern can be reviewed on its own.

## What changed

1. **File moves** (commit 1): 1,806 source files moved to the standard layout (`src/main/java`, `src/test/java`, …). `git show --stat -M100%` confirms every one is a pure move.
2. **The build swap** (commit 2): the new Gradle build via a committed, checksum-pinned wrapper. Everything lands where it did: `server/setup`, `server/dist`. The Ant entry points become stubs that point to `./gradlew`; the subproject Ant files are deleted. CI, the Dockerfile, and the contributor docs use the new build (with `-P` flags). Eclipse-era conveniences are now build commands (`devRun`, `devLauncher`, `devClient`, `createDerbyDb`).
3. **The dependency swap** (commit 3): the 374 verified jars come from Maven Central; the 45 unverifiable files (32 distinct libraries) stay vendored with per-jar reasons (appendix). Three safeguards: every download is checksum-verified on every build; the build fails loudly if a jar would silently go missing; and transitive resolution is kept off so the shipped set cannot drift (see Limitations).

## How mistakes are caught before they ship

The design rule everywhere: **a mismatch must stop the build with a clear message, never surface later as a broken product.**

- Update a library version but forget to refresh the fingerprints? The build fails before compiling, naming the library.
- A downloaded jar with no assigned place in the shipped product? The build fails and names it.
- The startup files that list jar names are generated from the resolved versions, so they can't go stale.
- Changing a library version is two steps, and forgetting either is a build failure with instructions (the exact command, and why it needs a cold cache, is in CONTRIBUTING).

For changes to the build logic itself, CONTRIBUTING documents a habit: snapshot the built product before and after, diff the two, confirm only what you intended changed. The tooling for that is `snapshot_distribution.py` / `compare_builds.py` in [oie-build-parity](https://github.com/pacmano1/oie-build-parity).

## How this was built

Substantial parts of this migration were produced with AI assistance: the Gradle build scripts, the comparison and provenance tooling, the dependency audit, and the drafts of this write-up. The experiment: can AI-driven work of this size be held to machine-checked claims rather than taken on trust? The controls:

1. **The pass/fail bar was set before the work began**: the new build's output must match the old one, verified by machine. Every aggressive step happened behind that gate, and it caught real mistakes (ledger below).
2. **Machine claims were treated as unproven until re-executed.**
3. **Every internal review ran its findings through a second, adversarial pass** whose job was to disprove them, a fresh AI session with none of the first's context, re-running the commands. (The ledger shows one at work: the disproven "zero deprecation warnings" claim.)
4. **Independent tooling.** The primary comparison tool was itself written as part of this AI-assisted work, so passing it is partly self-grading. The output was additionally verified with [oie-release-verifier](https://github.com/OpenIntegrationEngine/release-verifier), the release-checking toolkit @mgaffigan wrote for a different purpose.
5. **A human approved every irreversible step.** Approving commands is not code review, and neither is the output-equivalence proof; the build logic still needs human eyes, and the three-commit structure exists for that.
6. **Discretionary improvements were kept out, even provably safe ones.** This change rests on one claim, nothing changed, and every optional improvement folded in makes it harder to verify. (Example: thirteen vendored jars are content-identical to their public copies and could safely become downloads; left untouched here, queued as follow-ups.)

## The error ledger

The mistakes made during this work, and what caught each:

- An early wrong AI guess blamed a compiler path setting for output differences; the real cause was a compiler feature silently disabled by Gradle's defaults. The output comparison caught it.
- The AI wrongly classified three libraries as "not on Maven Central." Re-execution caught it: Central's *search index* is missing fingerprint records for some files, so "no match" must be confirmed against the actual download server.
- An AI refactor dropped eleven historically renamed jars from the shipped product. The output comparison flagged all eleven.
- An internal AI review claimed "zero deprecation warnings." A fresh cold review re-ran the command and disproved it; the build was restructured. One further-out item remains (Limitations).
- The AI's first boot-test ran beside a live server instance, so its result couldn't be attributed to the new build. A human caught it; runtime testing moved to a human-run boot, which succeeded.
- The dependency checksum inventory was first generated on a warm cache without CI's coverage flag, missing parent-POM and JaCoCo-agent entries. The fail-closed gate stopped CI within seconds; it was regenerated cold with CI's flags, and CONTRIBUTING documents that.
- Project lore held the HAPI 2.3 jars were locally patched. A human question prompted an entry-by-entry comparison: content-identical to the public copies, only repackaged.

## Known limitations

- **Adding a new library takes more manual work** (transitive resolution is off, so the shipped set provably can't drift). The way out is gradual: when a library is upgraded anyway, switch transitivity on for just that one and check with the comparison tool.
- **The build logic is centralized and plain, not idiomatic.** Large Groovy files transcribing the old Ant build one-to-one, guarded by the output comparison rather than unit tests. Restructuring into convention plugins is left for its own change.
- **The configuration cache isn't supported yet.** One call is slated for removal in Gradle 10 (years out); the build pins Gradle 8.14.1 and is unaffected. Tracked.
- **Signed release builds are ~5 minutes slower** (signing one jar at a time vs four). Only signed release builds notice.
- **Of the three commits, only the last builds.** Split this way so each part can be *read* in isolation.
- **A few `…Tests`-named classes have never run**, before or after (the pattern matches `…Test`). Preserved unchanged; a separate follow-up.
- **The equivalence proof covers the cutover only.** What protects every day after: per-build checksum verification, the fail-closed placement gate, the 654 tests, and the documented snapshot habit.
- **Output claims are verifiable; process claims are not.** Every claim about the build's output can be re-run. The claims about how the work was done can't be checked, and don't need to be: a reader who ignores the process story entirely can verify the build the same way.

## Appendix: the 45 jars that stay in the repository

Generated from the committed audit record (`jar-provenance.json` in [oie-build-parity](https://github.com/pacmano1/oie-build-parity)). 32 distinct jars, 45 files across modules.

- **Same contents as the public copy, different wrapper (13):** unpacked and compared file-by-file against Maven Central; every file inside identical, but the zip container fingerprints differently (packing order/timestamps/compression). Strict file-level identity was the adoption rule, so these stayed. Safe to adopt later. `autocomplete-2.5.4`, the ten `hapi-*-2.3` jars, `jtds-1.3.1`, `rsyntaxtextarea-2.5.6`, `sqlite-jdbc-3.43.2.1`.
- **Genuinely changed from the public copy (3):** compiled code differs (114 classes in `javaparser-1.0.8`, 60 in `zip4j_1.3.3`, 2 in `not-going-to-be-commons-ssl-0.3.18`). Local forks; replacing them would change the product.
- **Nothing on Maven Central to replace them with (10):** `PDFRenderer`, `backport-util-concurrent-Java60-3.1`, `istack-commons-runtime-3.0.6`, `jai_imageio`, `language_support`, `looks-2.3.1`, `openjfx`, `webdavclient4j-core-0.92`, `wizard`, `wsdl4j-1.6.2-fixed`.
- **Known origins outside Maven Central (6):** five `dcm4che-*-2.0.29` jars (from dcm4che.org) and `mirth-vocab.jar` (built in-house by the `generator` module).

## How to re-verify

Needs JDK 17, Apache Ant (acceptance used 1.10.14), Python 3, and network access for the first Gradle run. The tooling lives in [oie-build-parity](https://github.com/pacmano1/oie-build-parity).

Run these from your checkout of this PR's branch (the first line captures it; the rest use absolute paths, so cwd doesn't matter after that):

```bash
ENGINE="$(pwd)"

# The comparison tooling
git clone https://github.com/pacmano1/oie-build-parity /tmp/oie-build-parity

# Ant baseline at the migration's parent commit, in a throwaway worktree
git worktree add /tmp/oie-ant 8c1111ba3
(cd /tmp/oie-ant/server && ant -f mirth-build.xml -DdisableSigning=true -DdisableTests=true build)

# Gradle build of this branch (same JDK build, same calendar day:
# version.properties embeds the build date). Gradle flags are -P;
# the Ant baseline above uses Ant's own -D syntax.
./gradlew clean build dist -PdisableSigning=true

# Compare every archive, entry by entry (absolute paths; cwd-independent)
python3 /tmp/oie-build-parity/compare_builds.py /tmp/oie-ant/server/setup "$ENGINE/server/setup"

git worktree remove --force /tmp/oie-ant
```

Expected: `REAL differences (0)`, with the known tool-fingerprint deltas (Ant-version labels, timestamps, `version.properties`, the two directory entries, and the empty `package-info.class` markers) each classified and counted. The Ant baseline builds in seconds (normal); the Gradle build takes a few minutes and runs the full suite.

A successful Gradle build ends with Gradle's stock "incompatible with Gradle 9.0" banner; `--warning-mode all` shows it refers to the one disclosed Gradle-10 item under Limitations.

The dependency audit is re-checkable with `python3 /tmp/oie-build-parity/sweep_provenance.py --verify-exact` (expect `0 mismatches`). The signed-build check and the rc1 cross-check each have their own recipe in the oie-build-parity README.

The Ant baseline hash file (oie-release-verifier format, uncompressed SHA-256 `e54cb2ef8c30ea887fb6752fc83a4dbf9fd818de3267177d29d6d2a1eaf3f6e4`) is published [here](https://github.com/pacmano1/engine/releases/download/ant-baseline-8c1111ba3/ant-baseline-hashes-for-pr.txt.gz). A second maintainer re-running the dual-build recipe before merge is invited.


mgaffigan:
Looking good!  Tests pass, CI passes, released binaries are byte-identical with the main ant build.

gradle/wrapper/gradle-wrapper.jar
jonbartels, line 1: Should this JAR be committed? What is typical for Gradle? Surely a depdendency manager can download its own jars?

jonbartels (changes requested):
Can the file moves be refactored to a separate PR. It is good to move the project to the standard gradle layout. However it would make the first commit smaller.  https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceSet.html and `main.java.srcDirs`

So that would be one PR with a non-standard layout. _then_ a second PR would move to the standard layout. It could also be one PR with multiple commits showing the addition of gradle, removal of JARs, then move of project files.

I think this PR has the same too-large problem one of the earlier Nico PRs had

.github/workflows/build.yaml
```diff
@@ -32,15 +32,18 @@ jobs:
           java-package: 'jdk+fx'
           distribution: 'zulu'
 
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@ed408507eac070d1f99cc633dbcf757c94c7933a # v4.4.3
+        with:
+          validate-wrappers: true
```
mgaffigan, line 38: What's the benefit of using this over the built-in wrapper?  Wrapper validation?

.gitignore
```diff
@@ -650,3 +650,10 @@ donkey/donkey-test
 # /server/src/org/mozilla/javascript/
 
 # /server/src/org/mozilla/javascript/xmlimpl/
+server/setup.tar.gz
+server/certchain.pem
+dependency-check-report.html
+dependency-check-report.json
+tools/install4j/oie-installer-config.install4j~
+yubikey-pkcs11.cfg
+certchain.pem
```
mgaffigan: These seem like they are unrelated.  I don't think it's a problem, but they don't seem strictly related to the gradle conversion.

BUILD-MIGRATION.md
mgaffigan, line 1: Seems like this should be part of the PR itself - not in-repo.  It's going to be stale quickly.  Remove.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+        assemblyEnabled = false
+    }
+    // also scan the jars that remain vendored (no Maven coordinates)
+    scanSet = subprojects.collect { it.file('lib') }.findAll { it.directory }
+}
+
+// -D and -P both work for the historical build flags.
+def flag = { String name ->
+    (System.getProperty(name) ?: providers.gradleProperty(name).getOrNull()) == 'true'
+}
```
mgaffigan, line 68: Should we maintain this?  Or take this as an opportunity for a breaking change.  `-D` to `-P` is not so big a lift for callers - easier to do so now than later.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+
+// Maps each Maven Central artifact ("group:artifact") to the directory
+// its vendored copy lived in, so the staged distribution layout matches
+// what the vendored build produced. Version-independent: bumping a
+// version in libs.versions.toml needs no layout change; only a NEW
+// distributed artifact needs a placement line (the build fails with a
+// clear message if one is missing).
+ext.vendoredLayout = file('gradle/vendored-layout.json').exists()
+        ? new groovy.json.JsonSlurper().parse(file('gradle/vendored-layout.json'))
+        : [:]
```
mgaffigan, line 90: Can we avoid this being a central json that is going to have merge conflicts?  Centralizing makes this unclear.  Each project (client/server) should specify the deps it requires.  Depending on details, might be better as a follow-up PR.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+        generatePackageInfoClasses(sourceSets.main.java.srcDirs, destinationDirectory.get().asFile)
+    }
+    compileTestJava.doLast {
+        generatePackageInfoClasses(sourceSets.test.java.srcDirs, destinationDirectory.get().asFile)
+    }
+
+    // The default java-plugin jar (build/libs/<name>.jar) is not an OIE
+    // artifact; every real jar is declared explicitly per module.
+    tasks.named('jar') { enabled = false }
+
```
mgaffigan: Let's remove this and push it to the normalization script of the verifier to delete files matching this hash.

Writing the raw bytestream for a class file is icky.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+                }
+            }
+            def f = inventoryFile.get().asFile
+            f.parentFile.mkdirs()
+            f.text = groovy.json.JsonOutput.toJson(inventory)
+        }
+    }
+
+    tasks.register('verifyDistributionPlacement') {
+        description = 'Fails if any resolved runtime artifact lacks a distribution placement.'
```
mgaffigan, line 231: Presumably this can be omitted if each project handles dep resolution individually.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+        preserveFileTimestamps = false
+        reproducibleFileOrder = true
+    }
+
+    // Gradle-9-safe cross-project artifact sharing: each project resolves
+    // its OWN runtime configuration and writes an inventory file; consumer
+    // projects read the file instead of resolving foreign configurations.
+    def inventoryFile = layout.buildDirectory.file('placement/artifacts.json')
+    tasks.register('writeArtifactInventory') {
+        description = 'Writes the resolved external runtime artifacts as JSON for cross-project staging.'
```
mgaffigan: I don't understand what this is doing for the build.  Is this effectively a debug log?  Presumably should be removed.

build.gradle
```diff
@@ -0,0 +1,416 @@
...
+    dependsOn ':donkey:clean', ':server:clean', ':client:clean', ':command:clean', ':generator:clean'
+}
+
+// Writes one SHA line per file of the staged distribution to
+// build/distribution-snapshot.txt. The protocol for changing build
+// logic: snapshot, make the change, build, snapshot again, diff. Only
+// the changes you intended should appear (see CONTRIBUTING.md).
+tasks.register('snapshotDistribution') {
+    group = 'verification'
+    description = 'Writes a SHA-256 line per file of server/setup for before/after comparison of build changes.'
```
mgaffigan: Seems like this is duplicative of the tools/verifier and separate verifiers.  The build cannot check itself.  Remove/move to tools.

CONTRIBUTING.md
```diff
@@ -26,8 +26,52 @@ git checkout -b feature/your-feature-name
...
+./gradlew build snapshotDistribution -DdisableSigning=true
+cp build/distribution-snapshot.txt /tmp/before.txt
+# ... make your build-logic change ...
+./gradlew build snapshotDistribution -DdisableSigning=true
+diff /tmp/before.txt build/distribution-snapshot.txt
+```
+
+Only the changes you intended should appear. For archive-level analysis
+of a difference, use `tools/build-parity/compare_builds.py` on two
+setup trees.
```
mgaffigan: This seems excessive for the getting-started.  Remove to wiki.

gradle/wrapper/gradle-wrapper.jar
sbraconnier: The purpose of the `gradle-wrapper.jar` is to replace a standard Gradle installation. So someone could clone the repo without Gradle being installed, and build the project. So it should be committed along the `gradle-wrapper.properties` file.

From the official [Gradle documentation](https://docs.gradle.org/current/userguide/gradle_wrapper.html):

> To make the Wrapper files available to other developers and execution environments, you need to check them into version control.
> 
> Wrapper files, including the JAR file, are small. Adding the JAR file to version control is expected. Some organizations do not allow projects to submit binary files to version control, and there is no workaround available.

tools/build-parity/README.md
mgaffigan, line 1: I don't think this directory landing in main is critical to this PR landing, I suggest the build-parity directory be removed to a separate PR if it is to be in main.

tools/install4j/oie-installer-config.install4j
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <install4j version="11.0.3" transformSequenceNumber="11">
-  <application name="Open Integration Engine" applicationId="4145-9206-7630-8076" mediaDir="${compiler:installer:mediaRoot}/server/build" shortName="oie" publisher="Open Integration Engine Project" publisherWeb="https://openintegrationengine.com" version="4.6.0" backupOnSave="true" autoSave="true" macVolumeId="88477e584eb462ba" javaMinVersion="17" javaMaxVersion="25">
+  <application name="Open Integration Engine" applicationId="4145-9206-7630-8076" mediaDir="${compiler:installer:mediaRoot}/server/installer" shortName="oie" publisher="Open Integration Engine Project" publisherWeb="https://openintegrationengine.com" version="4.6.0" backupOnSave="true" autoSave="true" macVolumeId="88477e584eb462ba" javaMinVersion="17" javaMaxVersion="25">
```
mgaffigan: Why is this changing?

mgaffigan (changes requested):
Generally looks really good.  (Obviously runs fine, since it is byte identical)

Changes requested:
- remove documentation of the change that will quickly grow stale (consider moving to github wiki or another destination)
- remove verifier to separate repo to avoid having to review
- remove/clarify purpose of apparently unused steps
- remove hardcoded "empty" class file generation.  Check in the .class file or remove and consider this an intended difference.

Other incidentals:
- Consider having the dependency resolution per-project file (unless it makes this easier to review, we can do that in a separate PR)
- Consider omitting the -D legacy support.  I don't think this will work well long term, and it is easier to make a breaking change now.

pacmano1:
Thanks, this was a useful pass. All of it is addressed; rundown:

**Docs that go stale**: `BUILD-MIGRATION.md` is removed; the write-up now lives in the PR description.

**Verifier to a separate repo**: done. `compare_builds.py`, the provenance audit, and the snapshot tool now live in https://github.com/pacmano1/oie-build-parity, out of this PR's review surface. If you'd rather it sit under the org, I'll move it to `OpenIntegrationEngine/oie-build-parity`; it's offered, not imposed.

**Unused/debug steps**: `snapshotDistribution` is removed (the snapshot protocol moved to the external tooling). `writeArtifactInventory` isn't a debug log; it records each module's resolved jars so `:server:stageSetup` can place foreign-module jars without resolving their configurations cross-project (the Gradle-9-safe path). I reworded its description so that's clear.

**Empty `package-info.class` generation**: gone. The build no longer hand-writes those three inert markers; they're now a documented, zero-runtime-effect difference, and the comparison tool classifies an ant-only empty `package-info.class` as expected. No raw bytestream in the build anymore.

**Per-project dependency resolution**: agree it's the better shape, and I'd like to do it as the follow-up you suggested, one step at a time with the parity check, rather than reshape staging inside the migration. The deps are already declared per-project (the catalog bundles); what's central is the placement map, which is version-independent and has had zero churn since it was created. Happy to take it on next.

**Drop `-D` legacy support**: done, clean break, no shim. Flags are `-P` now (`-PdisableSigning`, `-Pcoverage`, `-Pversion`, and `-Pkey.storepass` for signing). CONTRIBUTING and CI updated.

**Wrapper action (build.yaml)**: you guessed it: `validate-wrappers: true` checks the committed `gradle-wrapper.jar` against Gradle's published known-good checksums (the supply-chain answer to "should we commit the jar"), plus the action wires up Gradle's dependency caching across runs.

**install4j / .gitignore incidentals**: the install4j change is reverted out of the PR (it was an unrelated local edit). The unrelated `.gitignore` entries are removed.

One thing worth flagging in good faith: dropping `-D` did break a caller I'd missed at first, the `Dockerfile` (`ARG GRADLE_BUILD_ARGS="-DdisableSigning=true"`), since a stale `-D` is silently ignored rather than rejected. It's fixed to `-P` in the same commit as the flag change. Called out because it's exactly the "does this break a caller" question your comment raised, and the answer was yes, now handled. I fixed the flag but didn't run `docker build` to exercise the container itself (it isn't CI-gated); the change just mirrors the unsigned Gradle build, which is verified green.

pacmano1:
Fair concern, and the diff is genuinely large. The reason it's one change rather than a sequence is that the evidence is one piece: every archive in the distribution is verified byte-identical against an Ant build of the parent commit. An intermediate PR with a non-standard layout (or the build swap without the dependency swap) would be a state that was never verified and was never meant to ship, so splitting there would mean asking reviewers to sign off on something the proof doesn't cover.

The intended way to review this isn't to read all the files; it's to run the comparator and watch it report zero real differences. The recipe is in the PR description, and the tooling is in a separate repo now so it isn't part of what you have to read here. That's the review surface, not the line count.

On the `srcDirs` idea specifically: pointing Gradle at the old layout would keep a non-standard layout that a second PR then has to undo, so it adds a step rather than removing one, and the move is the cheap part to review anyway since `git show --stat -M100%` proves it's pure renames with zero content change.

The branch is already three commits split along the lines you describe: the 1,806 pure file moves (that same `-M100%` rename check), then the build swap, then the dependency swap, so each concern can be read on its own, and squash-on-merge is fine. Happy to walk through any one of them.

Also, thanks for the Windows SDKMan note, I rebased to address the review and kept your commit on top.

jonbartels (dismissed):
TY for splitting the commits up.

mgaffigan (dismissed):
We definitely need some follow-up PR's to split out dependency resolution, plugins, server, and client.  I think this is a very definite step in the right direction, though.

Verified with zulu17.60.17-ca-fx-jdk17.0.16-macosx_aarch64 against 88c8ee5d75b7c265a783f1decea3c14e67b3414d using https://github.com/mgaffigan/oie-release-verifier.  [Results were clean](https://github.com/user-attachments/files/29478370/88c8ee5.vs.64e1a98.patch).

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,8 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionSha256Sum=845952a9d6afa783db70bb3b0effaae45ae5542ca2bb7929619e8af49cb634cf
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.1-bin.zip
```
jbeckers, line 4: Probably best to use the current version https://gradle.org/releases/#9.6.1

.gitattributes
```diff
@@ -0,0 +1,13 @@
+# Line-ending rules for files where endings are functional.
+# (Pattern follows PR #214 by NicoPiel.)
+
+# Unix start scripts must stay LF
+/gradlew        text eol=lf
+oieserver       text eol=lf
+configure-from-env text eol=lf
```
jbeckers, line 7: Unrelated to this PR, but a very welcome correction. This bit me last week!

.gitignore
jbeckers, line 142: This and below comments/headers should point to src/main/java

CONTRIBUTING.md
```diff
@@ -26,8 +26,30 @@ git checkout -b feature/your-feature-name
...
 ### 5. Install Tooling
-OIE specifies the working versions of Java and Ant in [.sdkmanrc](./.sdkmanrc). To take advantage of this, install [SDKMAN](https://sdkman.io/) and run `sdk env install`
-in the project's root directory.
+OIE specifies the working Java version in [.sdkmanrc](./.sdkmanrc). To take advantage of this, install [SDKMAN](https://sdkman.io/) and run `sdk env install`
+in the project's root directory. No other tooling is required.
+
+### 5a. Build
+The build is driven by Gradle through the included wrapper; no separate Gradle install is needed:
+```bash
+./gradlew build -PdisableSigning=true   # full build without jar signing
```
jbeckers, line 35: Follow-up: should disableSigning be true by default?

jbeckers (changes requested):

jbeckers:
Looks good, some minor comments.

Can't wait for this to land!

NicoPiel:
Is there a reason this wasn't done it Kotlin? It is Gradle's official recommendation.

generator/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+    source = fileTree('vocab/src')
+    destinationDirectory = file('vocab/classes')
+    classpath = sourceSets.main.runtimeClasspath
+    options.encoding = 'UTF-8'
+    dependsOn generateVocabSource
+}
+
+tasks.register('vocabJar', Jar) {
+    description = 'Builds the mirth vocabulary jar from generated sources.'
+    archiveFileName = 'mirth-vocab-1.2.jar'
```
NicoPiel, line 42: Do we want a fixed version number here? I know it's there now, but it seems a weird thing to do.

gradle/wrapper/gradle-wrapper.properties
NicoPiel: Seconded.

CONTRIBUTING.md
NicoPiel: I think that's okay. Nobody signs JARs during development; signing explicitly better represents how we do things, imho.

NicoPiel (dismissed):
A few non-blocking comments. LGTM otherwise. Definitely something we can build on.

gradle/wrapper/gradle-wrapper.properties
mgaffigan: Gradle 9 drops support for all JREs below 17. Gradle 8 lets someone with any halfway decent JVM as JAVA_HOME build without issue. 

I agree with getting this to current, but @jbeckers are you ok with doing this in a later PR?

CONTRIBUTING.md
mgaffigan: This matches existing convention (primarily to allow testing with MCAL, which requires signing). I agree with changing the default, but would request we do so in a separate PR to avoid hanging up this one.

qpointsystems:
So great to see this. 15 years overdue for Connect, but better late than never, haha.

gradle/wrapper/gradle-wrapper.properties
jbeckers: I'm okay with shipping this with Gradle 8, I just thought it would be a good idea to at least try if this works with Gradle 9 as well.

On the other hand: "a halfway decent JDK" is at least 17 in my book (though I admit we're still migrating form 8 to 21 at work :) ).
And since .sdkmanrc specifies JDK 17 anyway, I'd guess everyone can get this version easily.

jbeckers:
> Is there a reason this wasn't done it Kotlin? It is Gradle's official recommendation.

I had the same thought, but as to not derail this effort, I kept the thought to myself. 😄 
Let's do that in a follow up.

pacmano1:
Thanks @jbeckers, ran through your notes:

- **Gradle version:** keeping the wrapper on 8.14.1 for this PR. Worth trying Gradle 9, but I don't want to fold a major-version bump into the migration. Tracking it as a follow-up.
- **.gitattributes:** glad it helps.
- **.gitignore headers:** those `# /client/src/com/...` lines are pre-existing (Nick Rupley's original 2017 .gitignore) and are inert comments, not active rules, so nothing's broken by them. Rather than churn 100+ comment lines into this diff, I'll clean them in the same follow-up.
- **disableSigning:** leaving the default as-is (signing on; pass `-PdisableSigning=true` to skip). That matches the Ant build exactly, which is the whole parity goal here.

Also synced with main via a merge commit (only new content is 3 of main's new files relocated onto the gradle layout) and CI is green. Would appreciate a re-review when you have a moment.



_Edit: the deferred follow-ups (Gradle 9, `.gitignore` cleanup, Kotlin DSL) are now tracked in #337._

mgaffigan (approved):

jbeckers (approved):
Looks good!

NicoPiel (approved):

kryskool (approved):

ssrowe (approved):

gibson9583 (approved):

.gitignore
pacmano1: Good catch. That commented block dates back to the original 2017 .gitignore and has never done anything, and it still references the pre-move layout. Rather than grow this PR's diff for inert comments, it's tracked in the follow-up list in #337 ("Clean up stale .gitignore headers"). Resolving for now.

generator/build.gradle
pacmano1: The fixed name matches what the Ant build produced. The old generator/build.properties had `vocab.version=1.2` and `vocab.jar=mirth-vocab-${vocab.version}.jar`, so the output was mirth-vocab-1.2.jar there too. I kept it literal so the parity check stays byte-identical against the Ant output. Agreed it reads oddly. Wiring it to a real version is a good candidate for the #337 follow-ups once we're off the parity constraint.

tonygermano (dismissed):
Can someone confirm that the verification tool has been run for the latest commit in this PR (d724b51ad) and the last ant commit (7e35abedb) since the merge was created?

mgaffigan:
> Can someone confirm that the verification tool has been run for the latest commit in this PR (d724b51ad) and the last ant commit (7e35abedb) since the merge was created? 

I verified with my tool before re-approving. Still identical.

qpointsystems:
Merged this into my fork this morning and what an amazing thing to see... gradle at last!!!  15 years and we finally made it!  

When do we deprecate ANT, haha.

Issue #335, opened 2026-07-02
[IDEA] Add Backend Support for a Web Administrator


gibson9583:
A web version of the client administrator is a nice feature to have, but there are additional endpoints required in the engine to _properly support_ a web client.

NicoPiel:
Like what?

gibson9583:
> Like what?


@NicoPiel : 
1.  Message tree serialization — the transformer/filter message trees need serialization for each data type. The Swing client gets this by calling the datatype serializers in-process; a browser can't.
2. Script tooling — the script editors validate and format JavaScript 
3.  Web-plugin delivery — An extension can ship a browser UI in a folder within the zip. The Swing client loads ClientPlugin classes from the jars, but  a web client needs the engine to serve that UI over REST, so a plugin's UI follows the engine it's installed on rather than installing on the web admin's filesystem.

PR #176, opened 2025-08-26
Refactor to allow CSRF bypass, improve error logging for API authoring

mgaffigan:
- Refactor to allow CSRF bypass (optional `X-Requested-With` header) on endpoints that opt-in with `@DontRequireRequestedWith`
- Improve error logging for API authoring on a few sharp edges

In support for future merge of [mgaffigan:feature/add-oidc-auth](https://github.com/OpenIntegrationEngine/engine/compare/main...mgaffigan:oie:feature/add-oidc-auth)

PR #336, opened 2026-07-02
feat: Add Endpoints Required to Support a Web Administrator (#335)

gibson9583:
Add engine REST support so a standalone web administrator can run fully against the engine by serving plugin UIs, exact serialization, and script tooling with no local JVM sidecar or shared filesystem.

Endpoints (all session-authed, no extra permission, auditable=false):
- GET  /api/webplugins                          — enabled extensions that ship a webadmin/ UI
- GET  /api/webplugins/{extensionPath}/{file}   — serve an extension's webadmin/ assets
- POST /api/javascript/_validate                — validate JS (Rhino compiler)
- POST /api/datatypes/_serialize                — serialize a message via a data type (message trees)

Supporting changes:
- RequestedWithFilter: exempt read-only GET /api/webplugins/* from the X-Requested-With CSRF requirement (browser module/script loads can't set headers).
- DataTypeServerPlugin: add getVocabulary(version, type); override in HL7v2, EDI, NCPDP, and DICOM so /api/datatypes/_serialize returns element descriptions.

New servlets auto-register via the existing package scan. No wiring changes.


server/src/com/mirth/connect/client/core/api/servlets/DataTypeServletInterface.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) Mirth Corporation. All rights reserved.
```
kayyagari: NG/Mirth header seems to be picked up by default, should be replaced with OIE license header.

gibson9583: all set. Removed

server/src/com/mirth/connect/server/api/servlets/DataTypeServlet.java
```diff
@@ -0,0 +1,260 @@
...
+
+            // Message type/version from the serializer's own metadata, then the data type's
+            // vocabulary (element descriptions) — the same text the Swing tree shows. Only the
+            // XML-serialized types decorate nodes; JSON is a plain object tree.
+            String[] tv = typeAndVersion(serializer, msg);
+            MessageVocabulary vocab = safeVocab(plugin, tv[1], tv[0]);
+            String root = buildRoot(tv[0], tv[1], vocab);
+            String descriptions = (!json && vocab != null) ? buildDescriptions(data, vocab) : "{}";
+
+            String out = "{"
```
kayyagari: Using ObjectNode from Jackson makes the JSON building clean. The `jsonString()` function becomes unnecessary afterwards.

gibson9583: Refactored this to use ObjectNode/Mapper

pacmano1 (changes requested):
We are on Gradle now, please rebase.

Needs tests. `RequestedWithFilterTest` already exists and you added a branch to it without adding a case. The traversal guard in `getWebPluginResource` is the one I actually care about.

Headers got deleted instead of replaced, so six files ship with no license at all. Add the SPDX pair like `AbstractXMLReader.java`.

mgaffigan:
I agree with adding the support endpoints (vocabulary, validate, ...).  Should you wish to reduce this PR to those only, I am happy to approve as is.

I do not understand the goal of avoiding WAR files.  WAR is the format for shipping static files.  This seems to be re-inventing a plugin-specific file serving concept.  

I would agree with having a way for plugins to deliver a manifest via API - that should be aggregated into a single merged document, though, not N requests to arbitrary URL's, though.

Should we have to serve static resources from plugin jars, then I would say that should be via https://github.com/OpenIntegrationEngine/engine/pull/176

pacmano1 (approved):
Traversal guard is covered properly now, and the headers are in on the servlets.

`RequestedWithFilterTest.java` is the one that got missed. No header at all, it starts straight at the package declaration. Add the SPDX pair and it's done.

Approving so this isn't waiting on me.

Issue #344, opened 2026-07-11
REST POST /api/channels accepts a structurally-invalid channel; it deploys "healthy" but sends corrupt (undefined) content


MichaelLeeHobbs:
# OIE bug: `POST /api/channels` accepts a structurally-invalid channel; it deploys "healthy" but sends corrupt content

**Component:** REST API — `ChannelServlet` `POST /api/channels`
**Version observed:** Open Integration Engine 4.5.2 (`openintegrationengine/engine:4.5.2-alpine`)
**Severity:** data-integrity — a channel that reports deployed and healthy silently sends corrupt messages.
**Scope:** reproduces on a **stock OIE with no extensions installed**, using the **built-in TCP Sender**.

## The issue

`POST /api/channels` accepts a channel whose XML is **structurally incomplete/invalid** — it neither
validates/rejects it nor normalizes it. The server then deploys it (channel and destination both `STARTED`,
no warnings), but the destination sends the message content as the literal string **`undefined`** on the
wire — a corrupt message — with no error anywhere.

Opening the same channel in the Administrator and saving it (which normalizes the config) fixes it. That
confirms the POSTed config was genuinely invalid **and** that the REST endpoint should have caught it — the
Administrator's create/save path validates/normalizes; the REST create path does neither.

## What "invalid" means here

Compared with the Administrator-normalized version of the same channel, the accepted XML has:

| Field | Accepted by REST | Administrator-normalized |
|---|---|---|
| `transmissionModeProperties class` | `com.mirth.connect.model.transmission.framemode.FrameModeProperties` (base) | `com.mirth.connect.plugins.mllpmode.MLLPModeProperties` (+ `useMLLPv2`/`ackBytes`/`nackBytes`/`maxRetries`) |
| `<pluginProperties/>` on each connector | *absent* | present (empty) |
| `queueBufferSize` | `0` | `1000` |

Correcting the transmission mode and `queueBufferSize` alone does **not** resolve the `undefined` output;
the Administrator's full normalization (which also adds `<pluginProperties/>`) does.

## Reproduce

Stock OIE 4.5.2, no extensions, plus any MLLP listener at the destination's `host:port`:

1. `POST /api/channels` with the attached **`stock-channel-incomplete.xml`** (Channel Reader source →
   built-in **TCP Sender** destination, template `${message.encodedData}`). → **HTTP 200 (accepted).**
[...]

PR #351, opened 2026-07-16
Fix silent message corruption from absent/blank channel scripts (#344)

e210:
Fixes the root cause behind #344: `POST /api/channels` accepts a channel without `<preprocessingScript>`; the channel deploys "healthy" but corrupts every message to the literal string `undefined` on the wire.

### Root cause

Two independent defects in `JavaScriptUtil` (full analysis in #344):

1. `JavaScriptBuilder.generateScript` string-concatenates a `null` script into the JS wrapper, producing `function doScript() { null }`. `compileAndAddScript` caches it as a "custom" preprocessor because its body differs from the default script.
2. `executePreprocessorScripts` guards results with `result != null`, but Rhino's `Undefined` is not Java `null` — `Context.jsToJava` coerces it to the string `"undefined"`, which becomes the processed raw and flows to every destination.

### Changes

- `compileAndAddScript`: a null/blank script is now treated as absent — evict from the compiled-script cache and return `false` (all callers already handle `false` as "no script").
- `executePreprocessorScripts`: both result checks (global + channel) now ignore `Undefined`, the same idiom `getPostprocessorResponse` already uses.
- New `JavaScriptUtilTest`: 5 tests exercising the real Rhino path (TDD — each fix developed against a failing test).

### Behavior change note

A user preprocessor with no `return` statement previously corrupted the message body to `"undefined"`; it now passes the message through unchanged. This is footgun-removal: returning the literal string `"undefined"` as content is never intentional (`return '';` and `return null;` behave as before).

### Verification

- Full `:server:test` suite green.
- E2E on the patched build (the issue's exact repro XML, MLLP sink logging raw bytes):

| Channel variant | Bytes on the wire (before) | Bytes on the wire (after) |
|---|---|---|
| Issue's incomplete XML as-is | `\x0b` `undefined` `\x1c\r` | intact HL7 message |

The broader ask in #344 — structural validation/normalization in `POST /api/channels` — is deliberately out of scope here; it deserves its own design discussion. This PR removes the data corruption.

### DIFF
--- a/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
+++ b/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
@@ -710,6 +710,13 @@ public class JavaScriptUtil {
         // Note: If the defaultScript is NULL, this means that the script should
         // always be inserted without being compared.
 
+        // A null or blank script does nothing; treat it as absent instead of
+        // compiling a wrapper whose body is the literal string "null" (MIRTH/OIE #344)
+        if (StringUtils.isBlank(script)) {
+            compiledScriptCache.removeCompiledScript(scriptId);
+            return false;
+        }
+
         boolean scriptInserted = false;
         String generatedScript = null;
 
--- a/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
+++ b/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
@@ -129,7 +129,7 @@ public class JavaScriptUtil {
             throw e;
         }
 
-        if (result != null) {
+        if (result != null && !(Undefined.isUndefined(result))) {
             String resultString = (String) Context.jsToJava(result, java.lang.String.class);
 
             if (resultString != null) {
@@ -197,7 +197,7 @@ public class JavaScriptUtil {
                 }
             }
 
-            if (result != null && !(result instanceof Undefined)) {
+            if (result != null && !(Undefined.isUndefined(result))) {
                 String resultString = (String) Context.jsToJava(result, java.lang.String.class);
 
                 // Set the processed message in case something goes wrong in the channel processor. Also update the global result so the channel processor uses the updated message
@@ -226,7 +226,7 @@ public class JavaScriptUtil {
                 }
             }
 
-            if (result != null && !(result instanceof Undefined)) {
+            if (result != null && !(Undefined.isUndefined(result))) {
                 String resultString = (String) Context.jsToJava(result, java.lang.String.class);
 
                 // Set the processed message if there was a result.
@@ -330,7 +330,7 @@ public class JavaScriptUtil {
                 // TODO: is it okay that we use Status.SENT here?
                 response = new Response(Status.SENT, object.toString());
             }
-        } else if ((result != null) && !(result instanceof Undefined)) {
+        } else if ((result != null) && !(Undefined.isUndefined(result))) {
             // This branch will catch all objects that aren't Response, NativeJavaObject, Undefined, or null
             // Assume it's a string, and return a successful response
             // TODO: is it okay that we use Status.SENT here?
--- a/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
+++ b/server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
@@ -197,7 +197,7 @@ public class JavaScriptUtil {
                 }
             }
 
-            if (result != null) {
+            if (result != null && !(result instanceof Undefined)) {
                 String resultString = (String) Context.jsToJava(result, java.lang.String.class);
 
                 // Set the processed message in case something goes wrong in the channel processor. Also update the global result so the channel processor uses the updated message
@@ -226,7 +226,7 @@ public class JavaScriptUtil {
                 }
             }
 
-            if (result != null) {
+            if (result != null && !(result instanceof Undefined)) {
                 String resultString = (String) Context.jsToJava(result, java.lang.String.class);
 
                 // Set the processed message if there was a result.
--- a/server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
+++ b/server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
@@ -17,15 +17,11 @@ import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import java.io.File;
-import java.lang.reflect.Field;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.util.HashMap;
 import java.util.HashSet;
 
 import org.junit.After;
-import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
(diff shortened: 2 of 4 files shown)


jonbartels (commented):
Holding my review until I hear from the original issue reporter - https://discord.com/channels/943670759891554316/943670760461987862/1528943487163044094

Looks good at a glance, but I'll let the OP verify it first

MichaelLeeHobbs:
Thanks for tracking this down — clean fix, and the root cause explains a symptom I couldn't (RAW channels corrupting too, since the preprocessor runs on the raw message before datatype handling). Nice test coverage on the real Rhino path.

server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
```diff
@@ -197,7 +197,7 @@ public static String executePreprocessorScripts(JavaScriptTask<Object> task, Con
                 }
             }
 
-            if (result != null) {
+            if (result != null && !(result instanceof Undefined)) {
```
tonygermano: This is the correct way to test for javascript undefined in java. There is one other incorrect usage in this file on line 333 which is not currently a part of this PR, but it would be great if you could fix that one, too, for consistency.

```suggestion
            if (result != null && !(Undefined.isUndefined(result))) {
```

server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
```diff
@@ -226,7 +226,7 @@ public static String executePreprocessorScripts(JavaScriptTask<Object> task, Con
...
```
tonygermano:
```suggestion
            if (result != null && !(Undefined.isUndefined(result))) {
```

jonbartels (dismissed):

NicoPiel (dismissed):

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        /*
+         * mirth.properties isn't on the unit test classpath, but JavaScriptScopeUtil's static init
+         * requires it to be resolvable via the context classloader. Point the context classloader
+         * at the real conf/ dir; restored after the class.
+         */
+        originalContextClassLoader = Thread.currentThread().getContextClassLoader();
+        URL confDir = new File("conf").toURI().toURL();
```
tonygermano: suggestion: I would prefer to see a copy of mirth.properties added as a test resource

Accessing Files directly from tests can be flaky depending on where the test is run from, and this file being referenced is not guaranteed to be static.

e210:
Checked both Copilot notes:

**Test mocks:** Mockito's default returns *empty collections*, not null — so no NPE, the loops just no-op. The 659 green tests in CI agree.

**Blank scripts (`DatabaseReceiverScript`):** fair, but tiny: a blank select script is an already-broken channel (the Administrator won't save it). Before it failed every poll with `DatabaseReceiverException("…: undefined")`, now with an NPE from the same `throw` line, caught by the same `catch`. Happy to add a null-guard to that error message in a follow-up if wanted.

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+        });
+        injector.getInstance(ControllerFactory.class);
+
+        /*
+         * JavaScriptBuilder captures its controllers in static fields at class-load time. If an
+         * earlier test class loaded it with a mocked factory that left them null, repair them so
+         * generateGlobalSealedScript/appendCodeTemplates don't NPE.
+         */
+        setJavaScriptBuilderStaticField("extensionController", extensionController);
```
tonygermano: suggestion: create package-private method for changing these values for testing

It seems a bit hacky to use reflection to change these values, but I understand why you did it. Refactoring JavaScriptBuilder to allow injectable dependencies is probably out of scope for this PR, but I think adding a method would be an acceptable compromise.

```java
// In JavaScriptBuilder.java (from Guava, which should already be on the classpath)
import com.google.common.annotations.VisibleForTesting;

@VisibleForTesting
static void setControllersForTesting(ExtensionController ec, CodeTemplateController ctc) {
    extensionController = ec;
    codeTemplateController = ctc;
}
```

tonygermano (changes requested):
See inline comments. Also, please respond to the co-pilot suggestions.

server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
e210: Applied in 42d4c1f, including line 333 — and a fourth occurrence of the same pattern at line 132 (attachment script result handling), fixed for the same consistency reason.

server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
e210: Applied in 42d4c1f.

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
e210: Done in 42d4c1f — mirth.properties copied to test resources, classloader shim removed entirely.

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
e210: Done in 42d4c1f — one tweak: the method had to be `public` rather than package-private, since the test lives in `server.util.javascript` while the builder is in `server.builders`. `@VisibleForTesting` marks the intent.

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
```diff
@@ -0,0 +1,184 @@
...
+        when(controllerFactory.createEventController()).thenReturn(eventController);
+
+        ConfigurationController configurationController = mock(ConfigurationController.class);
+        when(controllerFactory.createConfigurationController()).thenReturn(configurationController);
+
+        ExtensionController extensionController = mock(ExtensionController.class);
+        when(controllerFactory.createExtensionController()).thenReturn(extensionController);
+
+        CodeTemplateController codeTemplateController = mock(CodeTemplateController.class);
+        when(controllerFactory.createCodeTemplateController()).thenReturn(codeTemplateController);
```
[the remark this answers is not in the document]
e210: Mockito's default answer (`ReturnsEmptyValues`) returns empty collections — not null — for `getCodeTemplates()`/`getLibraries()`/`getPluginMetaData()`, so `appendCodeTemplates()` and `generateGlobalSealedScript()` no-op rather than NPE. The full suite (659 tests, real class ordering) runs green in CI. The `@BeforeClass` repair covers the ordering concern — now via the `setControllersForTesting` method added in 42d4c1f.

server/src/main/java/com/mirth/connect/server/util/javascript/JavaScriptUtil.java
```diff
@@ -710,6 +710,13 @@ public static boolean compileAndAddScript(String channelId, MirthContextFactory
         // Note: If the defaultScript is NULL, this means that the script should
         // always be inserted without being compared.
 
+        // A null or blank script does nothing; treat it as absent instead of
+        // compiling a wrapper whose body is the literal string "null" (MIRTH/OIE #344)
+        if (StringUtils.isBlank(script)) {
+            compiledScriptCache.removeCompiledScript(scriptId);
+            return false;
+        }
```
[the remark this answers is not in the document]
e210: True in principle, but the only affected configs were already broken: a blank select script cannot be saved from the Administrator (REST-only — which is how this whole issue started). Before this PR every poll threw `DatabaseReceiverException("…: undefined")`; now it is an NPE from the same `throw` line, caught by the same `catch`. Same loud failure either way. Scoping the behavior per-context and compiling empty wrappers elsewhere would reintroduce the phantom-script executions this PR removes. Happy to add a null-guard to that error message in a follow-up if wanted.

mgaffigan (approved):

server/src/test/java/com/mirth/connect/server/util/javascript/JavaScriptUtilTest.java
tonygermano: Apologies for making that suggestion and not noticing the classes were not in the same package. I'm not really happy with either solution, but I'll leave this one at "good enough."

tonygermano (approved):

NicoPiel (approved):

PR #356, opened 2026-07-17
Add Dependabot config (Gradle + Docker) with grouping and verification-metadata automation

jonbartels:
## Summary

Adds Dependabot dependency updates for this project, plus a workflow to keep them compatible with the repo's Gradle dependency verification.

- **`.github/dependabot.yml`** — config schema `version: 2`
  - **Ecosystems:** `gradle` (root `/`, covers the multi-module build via the central `gradle/libs.versions.toml` catalog) and `docker` (the root `Dockerfile`)
  - **Schedule:** `monthly` for both
  - **Grouping:** related artifacts are bundled into one PR, keyed at the *coherent-project* level. `org.apache.*` is deliberately split (`org.apache.commons`, `org.apache.logging.log4j`, …) because it spans unrelated projects (Derby, PDFBox, Velocity, HttpComponents). Groups: `aws-sdk`, `jetty`, `jersey`, `jackson`, `netty`, `log4j`, `apache-commons`, `test-tooling`, and a catch-all `all-minor-and-patch`
  - **Noise controls:** `open-pull-requests-limit` (10 gradle / 5 docker), `labels`, `commit-message` prefixes
- **`.github/workflows/dependabot-verification-metadata.yaml`** — Dependabot cannot update `gradle/verification-metadata.xml` ([dependabot-core#1996](https://github.com/dependabot/dependabot-core/issues/1996)), so its Gradle bumps would fail CI on stale checksums. On `dependabot/gradle/**` branches this workflow regenerates the metadata with a cold cache (per `CONTRIBUTING.md`), validates the build in the same run, and commits the refreshed file back to the PR.

## ⚠️ Manual configuration required (by a repo admin)

These cannot be done from the PR and are needed for the config to behave as described:

1. **Create the labels** (Dependabot only auto-creates its own defaults; custom labels that don't exist are silently dropped, so PRs would arrive unlabeled):
   ```sh
   gh label create dependencies -c "#0366d6" -d "Dependency updates" --repo OpenIntegrationEngine/engine
   gh label create java         -c "#b07219" -d "Java / Gradle dependencies" --repo OpenIntegrationEngine/engine
   gh label create docker        -c "#1d63ed" -d "Docker base image updates" --repo OpenIntegrationEngine/engine
   ```
2. **Enable Dependabot version updates** for the repo if not already on: *Settings → Code security → Dependabot → Dependency graph + Dependabot version updates.* The config file only governs *how* updates run once the feature is enabled.
3. **Give Actions write permission** so the metadata workflow can push its commit: *Settings → Actions → General → Workflow permissions → "Read and write permissions."*
4. **(Recommended) Make the metadata workflow a required status check** on Dependabot PRs via branch protection. A push made with the default `GITHUB_TOKEN` does not re-trigger `build.yaml`, so the regen workflow is the authoritative green check for Gradle bumps (it runs the full build itself). If you'd rather have `build.yaml` re-run, swap the workflow's push to use a PAT/App token instead of `GITHUB_TOKEN`.

## Notes / follow-ups

- **netty inline pin:** `server/build.gradle` pins `netty-transport-native-epoll` inline with a `linux-x86_64` classifier (the catalog can't express it), so a grouped `netty` bump won't touch that line — watch for drift on the first netty PR.
- A `github-actions` ecosystem entry would be a reasonable follow-up to keep workflow action versions current.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "gradle"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,74 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    # Wait 2 days after a release before opening a PR, to skip yanked/broken versions.
+    cooldown:
+      default-days: 2
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "java"
+    commit-message:
+      prefix: "chore(deps)"
+      include: "scope"
+    # Groups are keyed at the coherent-project level. Note e.g. org.apache.* is NOT a
+    # single project — org.apache.commons, org.apache.logging.log4j, org.apache.pdfbox,
+    # etc. are unrelated and must not be bundled together, so patterns go three levels deep.
+    groups:
+      aws-sdk:
+        patterns:
+          - "software.amazon.awssdk:*"
+          - "software.amazon.eventstream:*"
+      jetty:
+        patterns:
+          - "org.eclipse.jetty*"
+      jersey:
+        patterns:
+          - "org.glassfish.jersey*"
+          - "org.glassfish.hk2*"
+      jackson:
+        patterns:
+          - "com.fasterxml.jackson*"
+      netty:
+        patterns:
+          - "io.netty*"
+          - "com.typesafe.netty*"
+      log4j:
+        patterns:
+          - "org.apache.logging.log4j*"
+      apache-commons:
+        patterns:
+          - "org.apache.commons:*"
+          - "commons-*:*"
+      test-tooling:
+        patterns:
+          - "org.mockito*"
+          - "net.bytebuddy*"
+          - "org.objenesis:*"
+          - "org.hamcrest:*"
+          - "junit:*"
+      # Everything else: bundle only minor/patch so major bumps still arrive as
+      # individual PRs that get reviewed on their own.
+      all-minor-and-patch:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    cooldown:
+      default-days: 2
+    open-pull-requests-limit: 5
+    labels:
+      - "dependencies"
+      - "docker"
+    commit-message:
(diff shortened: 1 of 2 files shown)


mgaffigan (changes requested):
Can we get the dockerfile included, too?

Also: add [minimum package age](https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/) to limit supplychain attacks.  2d would be my suggested minimum age.

Wissididom:
If I understand the github changelog entry correctly 3 days is already the default: https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/

mgaffigan (dismissed):

gibson9583 (dismissed):

kryskool (approved):

pacmano1:
One problem nobody has raised: every Gradle PR this config generates will fail CI. This repo enforces dependency verification (`gradle/verification-metadata.xml`, `verify-metadata=true`) and Dependabot cannot update that file: dependabot/dependabot-core#1996, open since 2020. I tested it locally by bumping `commons-cli` 1.2 -> 1.5.0 in the catalog, the same one-line change Dependabot would push:

```
Dependency verification failed for configuration ':server:compileClasspath'
3 artifacts failed verification:
  - commons-cli-1.5.0.jar (commons-cli:commons-cli:1.5.0) from repository MavenRepo
  - commons-cli-1.5.0.pom (commons-cli:commons-cli:1.5.0) from repository MavenRepo
  - commons-parent-52.pom (org.apache.commons:commons-parent:52) from repository MavenRepo

BUILD FAILED in 856ms
```

The build fails at the first task that resolves the bumped artifact, so most or all of the test suite never runs. The fix per PR is the cold-cache regeneration step in CONTRIBUTING.md.

Not a blocker for merging, but two things come with it:

- The `dependencies`, `java`, and `docker` labels don't exist in this repo. Dependabot ignores custom labels that don't exist (it only auto-creates its defaults). They need creating or the PRs come in unlabeled.
- Until a workflow automates the metadata regen (`--write-verification-metadata sha256` with a cold cache on Dependabot branches), the monthly batch will be red. Reviewers should expect that; it's not a misconfiguration.

Three minor items, none blocking:

- The comment on `all-minor-and-patch` says majors arrive as individual PRs. Only true for ungrouped dependencies; the named groups have no `update-types` restriction, so a Jackson or Jetty major would be included in its group PR.
- `server/build.gradle` pins `netty-transport-native-epoll` inline (the catalog can't express the `linux-x86_64` classifier), so a grouped netty bump won't update that line. Check it for drift on the first netty PR.
- Title and body still describe the first revision: weekly, gradle-only, no groups. It's now monthly with Docker coverage and grouping. A `github-actions` ecosystem entry might be a useful follow-up.

mgaffigan (approved):

jonbartels:
I am closing this PR. Instead I am pursuing Renvoate under https://github.com/OpenIntegrationEngine/engine/pull/358 . 

Renovate seems to require fewer permissions and workarounds than dependabot (it updates the gradle metadata automatically).

PR #361, opened 2026-07-23
Prevent SQL injection in message metadata column search

pacmano1:
## Problem

Custom metadata column names in message search are interpolated into SQL as identifiers (MyBatis `${}`) across all database dialects — `${element.columnName}` (metaDataSearch) and `${column}` (textSearchMetaDataColumns) — and arrive from the REST parameters with no validation, so an authenticated `MESSAGES_VIEW` user can inject arbitrary SQL. The search value (`#{...}`) and operator (enum-constrained) were already safe; only the column name is affected. Ref GHSA-mf46-33w3-84j6.

## Fix

Validate every referenced metadata column name against the channel's own defined columns before the filter reaches the data layer:

- **`MetaDataColumnValidator`** (new, `server/util`): pure check, lazy channel-column lookup, returns the first unknown column name (or null).
- **`MessageServlet`**: primary gate at the REST boundary; an unknown column returns `400`. Applied to all message-search entry points, in both the parameter and `MessageFilter`-body forms (`getMessages`, `getMessageCount`, `reprocessMessages`, `removeMessages`, `exportMessagesServer`).
- **`DonkeyMessageController`**: defense-in-depth backstop for callers that reach the controller directly (e.g. plugins via `MessageController.getInstance()`), throwing on an unknown column.

Column names are matched exactly against the channel's (upper-cased) defined columns. No SQL/mapper changes; the value and operator are untouched.

## Tests

- `MetaDataColumnValidatorTest` — unit tests the check: unknown/known/non-uppercase/null column, and that the channel lookup is skipped when the filter references no custom column.
- `MessageServletTest` — REST-boundary reject/allow paths.
- Full clean `:server` build green (unit tests included).

## Verifying

Verified against a running server (Derby) with a channel that has a custom metadata column:

- `metaDataSearch=<definedColumn> = foo` → `200` (search runs normally).
- `metaDataSearch=<undefinedColumn> = foo` → `400` (rejected before reaching SQL).


server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
```diff
@@ -387,7 +401,30 @@ private void sendServerEventWithAttributes(Map<String, String> attributes) {
...
+        // Primary check at the REST boundary: reject any custom metadata column the filter references
+        // that is not defined on the channel. Column names are matched exactly against the channel's
+        // (upper-cased) columns, which come from the in-memory channel cache (no database hit) and are
+        // only looked up when the filter actually references a custom column. Callers that reach
+        // DonkeyMessageController directly (e.g. a plugin) are covered by a last-resort backstop there.
+        String unknownColumn = MetaDataColumnValidator.findUnknownColumn(filter, () -> channelController.getMetaDataColumns(channelId));
+        if (unknownColumn != null) {
+            logger.warn("Rejected message search for channel " + channelId + " referencing unknown metadata column: " + unknownColumn);
+            throw new MirthApiException(Response.Status.BAD_REQUEST);
+        }
```
mgaffigan: Do we need to do this twice?  Seems extra and easy to have bugs at one of the two layers.  I suspect the goal was a 400 instead of a 500 - a typed exception might be a better route for that.

server/src/main/java/com/mirth/connect/server/controllers/DonkeyMessageController.java
```diff
@@ -295,6 +298,7 @@ public List<Attachment> getMessageAttachment(String channelId, Long messageId, b
 
     @Override
     public void removeMessages(String channelId, MessageFilter filter) {
+        validateMetaDataColumns(channelId, filter);
```
mgaffigan: We're having to "remember" to call this validator several times.  Is there an alternative that's harder to accidentally omit?  Seems like all entrypoints eventually lead to searchAll

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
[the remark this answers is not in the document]
mgaffigan: Replacing newlines is playing whack-a-mole, but this should be fixed to structured logging (`"{}"` in the log message format).

mgaffigan (dismissed):
It seems fine.  See below for notes if you feel like making edits.

server/src/main/java/com/mirth/connect/server/controllers/DonkeyMessageController.java
```diff
@@ -676,6 +681,21 @@ private List<MessageSearchResult> searchMessages(MessageFilter filter, String ch
...
+     * search mappers interpolate these column names into SQL as identifiers, so an undefined column
+     * must never reach them. This throws (rather than returning a 400) because reaching it means a
+     * caller skipped the proper gate, which is a programming error.
+     */
+    private void validateMetaDataColumns(String channelId, MessageFilter filter) {
+        String unknownColumn = MetaDataColumnValidator.findUnknownColumn(filter, () -> ControllerFactory.getFactory().createChannelController().getMetaDataColumns(channelId));
+        if (unknownColumn != null) {
+            throw new IllegalArgumentException("Message search referenced a metadata column that is not defined on channel " + channelId + ": " + unknownColumn);
+        }
+    }
```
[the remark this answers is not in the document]
mgaffigan: This is whack-a-mole.  Users should use structured logging to guard against log injection.

pacmano1:
Fair points.

Single gate: simplifies things. One caveat on reprocess: it's async, so a bad request still gets a response, but the call itself fails and logs an error.

Logging: will update, but it's not a perfect fix. I'll file a separate issue with the details.

Null elements: the null-entry NPE Copilot caught turns a bad request into a 500. Fixing it.

pacmano1:
Reworked since your review. What changed and why.

**Where the check lives:** moved out of the per-endpoint calls into the `FilterOptions` constructor. Every search path (get, count, remove, reprocess) builds `FilterOptions` once before its batch loop, so it's a single choke point a new endpoint can't skip, and it also covers a plugin calling the controller directly. Not in `searchAll`: that runs inside the batch loop, which is skipped when the message id range is empty (empty channel, or a min/max filter that excludes everything), so a bad column would slip past there without a query running.

**Status on rejection:** the check throws a plain `MetaDataColumnException` that bubbles up to the engine's standard 500, same as any other uncaught controller exception. I looked at a clean 400 and decided against it. The engine has no central exception-to-status mapping; every specific status is thrown explicitly from a servlet. Getting a 400 would mean either making this data-layer exception extend `WebApplicationException` (the only HTTP-carrying exception below the servlet layer anywhere in the tree) or adding per-endpoint catches (back to the placement you flagged). Not worth it: the injection is blocked either way since the throw is before any query runs, and a 500 is how the engine already surfaces every other controller-layer rejection.

**Two notes on the 500:** the exception message is generic, so the untrusted column name is not reflected back. The body does include a stack trace, but that's the engine's standard 500 behavior for every endpoint, not something introduced here.

**Live-tested on a running engine** (populated channel):
- `count/_search`, `_search`, `_remove` with an injected column -> 500, nothing deleted on remove
- legit column -> 200
- out-of-range message id filter + injection -> 500 (choke still fires)
- reprocess with injection -> 204, run aborted and logged, nothing reprocessed

Validator logic is unit-tested (`MetaDataColumnValidatorTest`); the exception type by `MetaDataColumnExceptionTest`.

**On verification vs. the current base:** the boot test above was run before syncing this branch with `main`. The current `main` won't start the engine because of the log4j lib mismatch that #389 fixes (unrelated to this change), so I'll re-run the boot test once #389 lands. Build and unit tests are green on the synced branch.

mgaffigan (dismissed):
This looks great! Tests are a bit "extra", but certainly not a problem.

NicoPiel (dismissed):

server/src/main/java/com/mirth/connect/server/util/MetaDataColumnValidator.java
```diff
@@ -0,0 +1,85 @@
...
+     * channel, or {@code null} if every referenced column is valid. Column names are matched exactly
+     * against the channel's (upper-cased) column names; a {@code null} referenced name is treated as
+     * unknown and returned as the string {@code "null"} so the result stays unambiguous.
+     *
+     * <p>
+     * The defined columns are supplied lazily and are only requested when the filter actually
+     * references a custom column, so a search that uses none costs no channel lookup.
+     * </p>
+     */
+    public static String findUnknownColumn(MessageFilter filter, Supplier<List<MetaDataColumn>> definedColumnsSupplier) {
```
tonygermano: As a new public method, I'd prefer to see this return an Optional instead of requiring explicit null handling by the caller. The javadoc should include `@param` and `@return` tags.

pacmano1: Done in 2daa835. Returns Optional<String> now, javadoc has the tags. Also fixed the class-level doc while I was in there, it still described the old two-layer REST/controller design. Build and unit tests green; the boot re-test is still parked behind #389.

mgaffigan (approved):

NicoPiel (approved):

gibson9583 (approved):

PR #388, opened 2026-07-24
Update renovate configs from maintainer and user feedback

jonbartels:
Fixes to address comments from users and maintainers on Renvoate bot.

https://github.com/renovatebot/renovate/discussions/35165 I think this is key. We reference reccomended presets but I think some of those generate behaviors that are undesirable to OIE. 

This PR was done by human hands and eyes, no clankers. I read https://docs.renovatebot.com/configuration-options/ and took the comments from maintainers and users in Discord. 

Grouping _is_ working but renovate submits security patches separately. 

The all-minor-and-patch rule needs to go. Theres like 50 packages in there. Thats not what we want. It bundles all those minor updates into one massive set. Pauls clanker called this out as non-blocking in PR #358 but I think it is a problem. Thats one big-big PR.

> If you're tweaking, I think the 2-day cooldown is too short, and something like 7-14 would be more appropriate for non-security patches. We don't move that quickly, and with it only checking once a month, anyway, we don't need it bringing in patches that fresh.

I can do days, but I do not think I can distinguish between security updates and other updates. This PR sets it to 14 days. I think security always gets pushed. See next topic.

> I know when I was looking, there is also a way to limit the number of PRs it has opened at one time, so that we don't get overwhelmed with them.

We have this. Its set to 10, but per Renvoate docs:
> Renovate always creates security PRs, even if the concurrent PR limit is already reached. Security PRs have [SECURITY] in their PR title.

The batch from PR #358 was entirely security updates. Renovate has many held updates that are non-security. These are visible at https://developer.mend.io/

Screenshot shows the ugly `all-minor-and-patch` behavior. It also shows groups working for non-security updates.
image

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -47,20 +47,20 @@
     {
       "groupName": "jetty",
       "matchPackageNames": [
-        "org.eclipse.jetty:*"
+        "org.eclipse.jetty*"
       ]
     },
     {
       "groupName": "bouncycastle",
       "matchPackageNames": [
-        "org.bouncycastle:*"
+        "org.bouncycastle*"
       ]
     },
     {
       "groupName": "jersey",
       "matchPackageNames": [
-        "org.glassfish.jersey:*",
-        "org.glassfish.hk2:*"
+        "org.glassfish.jersey*",
+        "org.glassfish.hk2*"
       ]
     },
     {
@@ -72,26 +72,26 @@
     {
       "groupName": "sun-xml",
       "matchPackageNames": [
-        "com.sun.xml:*"
+        "com.sun.xml*"
       ]
     },
     {
       "groupName": "jackson",
       "matchPackageNames": [
-        "com.fasterxml.jackson:*"
+        "com.fasterxml.jackson*"
       ]
     },
     {
       "groupName": "netty",
       "matchPackageNames": [
-        "io.netty:*",
-        "com.typesafe.netty:*"
+        "io.netty*",
+        "com.typesafe.netty*"
       ]
     },
     {
       "groupName": "log4j",
       "matchPackageNames": [
-        "org.apache.logging.log4j:*"
+        "org.apache.logging.log4j*"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


mgaffigan (dismissed):

NicoPiel (dismissed):

tonygermano (changes requested):
The PR comments claim that minimumReleaseAge on line 10 was adjusted to 14 days, but it is still at 2 days.

According to [the docs](https://docs.renovatebot.com/configuration-options/#schedule), the schedule on line 7 is using a deprecated syntax. It suggests using cron syntax or one of the presets.

jonbartels:
> The PR comments claim that minimumReleaseAge on line 10 was adjusted to 14 days, but it is still at 2 days.
> 
> According to [the docs](https://docs.renovatebot.com/configuration-options/#schedule), the schedule on line 7 is using a deprecated syntax. It suggests using cron syntax or one of the presets.

Fixed.

The minimumReleaseAge was updated locally but not pushed.

The schedule was updated to use a preset. I wasn't sure if renovate wants `monthly` or a fully qualified `schedule:monthly` for the preset value though.

mgaffigan (dismissed):

mgaffigan:
While we're here, you might also set `"rebaseWhen": "conflicted"` to avoid some of these PR updates every time main changes.

mgaffigan (dismissed):

NicoPiel (dismissed):

renovate.json
```diff
@@ -4,13 +4,14 @@
     "config:recommended"
   ],
   "schedule": [
-    "before 6am on the first day of the month"
+    "monthly"
```
tonygermano: I think if you are using a preset instead of cron that you need to remove the `schedule` section and add `"schedule:monthly"` to the `extends` section.

See https://docs.renovatebot.com/key-concepts/presets/#how-to-use-presets

tonygermano (changes requested):

renovate.json
jonbartels: It's sort of like the presets merge in with my renovate.json rather than used piece by piece

mgaffigan (approved):

kryskool (approved):

kpalang (approved):

NicoPiel (approved):

tonygermano (approved):

PR #389, opened 2026-07-25
Fix log4j bad versions

mgaffigan:
main (ac47cc88fc4d1c7efa4b79bed01236b6f141ac2d) has a bad combination of log4j libs which prevents the server from booting.  They were not caught on #370 since #294 has not yet merged to start the server as part of CI, so approvers got happy green checkmarks when everything is broken.

* Updates log4j to matching versions
* Fixes secondary log4j discovery issue
* Updates log4j to current latest

Tested and passed with #294 and through manual tests.


pacmano1 (approved):

NicoPiel (approved):

gradle/verification-metadata.xml
```diff
@@ -92,6 +92,11 @@
             <sha256 value="ba6d68eeab3a1cc13a77a8ade2197ff9a32aa1cffeac72847d49badd82e1b9ce" origin="Generated by Gradle"/>
          </artifact>
       </component>
+      <component group="com.fasterxml.jackson" name="jackson-bom" version="2.19.2">
```
tonygermano, line 95: Not sure what's going on with the jackson stuff. I don't think it is supposed to be in this PR? Additionally, this version doesn't match what is in libs.versions.toml, and there are already newer versions specified in this file.

mgaffigan: This was generated by the normal Gradle regen command. I assumed it had some interaction with another package or was missed in a prior PR.

jonbartels (approved):

gradle/verification-metadata.xml
mgaffigan: @tonygermano, is this blocking or are we good to merge?

mgaffigan: @tonygermano, main is still broken. We have enough reviews. I'm going to merge tomorrow evening unless you declare this blocking.

tonygermano: I won't say it's blocking, but I'd prefer to see the jackson stuff removed from this PR. Whatever it's doing is not enough to fix the remaining jackson issues.

pacmano1: They're log4j's. The parent POM's jackson-bom import moved 2.19.1 -> 2.19.2 with the bump, and 2.19.2's parent is jackson-parent 2.19.3. Removing both fails the build:

```
> Dependency verification failed for configuration ':donkey:runtimeClasspath'
  2 artifacts failed verification:
    - jackson-bom-2.19.2.pom (com.fasterxml.jackson:jackson-bom:2.19.2)
    - jackson-parent-2.19.3.pom (com.fasterxml.jackson:jackson-parent:2.19.3)
```

PR #390, opened 2026-07-25
Speed up test runs to ~30s (from 150s+)

mgaffigan:
TcpDispatcherTest was doing lots of sleeping to for synchronization.  Avoid that by using latches and other synchronization.  Also: avoid 9s of waiting for the 111.1.1.1 IP to be bindable.  Saves 90s or so.

TcpRecieverTest had another 10s of sleeps.

PropertiesConfig had another 7s.

Enabling parallel execution gave additional 20s+ back.

No expected behavior changes nor gaps in coverage.  Tested in a loop for stability:

```bash
while JAVA_HOME="/Users/mgaffigan/dev/zulu17.60.17-ca-fx-jdk17.0.16-macosx_aarch64" ./gradlew build -PdisableSigning=true && git clean -fdx; do :; done
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -264,6 +264,11 @@ subprojects {
         maxHeapSize = '2048m'
         jvmArgs '-Xshare:off'
         include '**/*Test.class'
+        /*
+         * Past ~4 forks the per-fork JVM startup cost (inflated by -Xshare:off above) outweighs the
+         * gain, since only a handful of classes take more than a second. Override with -PtestForks=N.
+         */
+        maxParallelForks = (project.findProperty('testForks') ?: Math.max(1, Math.min(4, Runtime.runtime.availableProcessors().intdiv(2)))) as int
     }
 
     if (rootProject.coverageEnabled) {
--- a/server/src/main/java/com/mirth/connect/client/core/PropertiesConfigurationUtil.java
+++ b/server/src/main/java/com/mirth/connect/client/core/PropertiesConfigurationUtil.java
@@ -28,6 +28,16 @@ import org.apache.commons.configuration2.reloading.PeriodicReloadingTrigger;
 
 public class PropertiesConfigurationUtil {
 
+    /**
+     * How long the reloading detector waits before it will stat the file again. This matches the
+     * default used by FileHandlerReloadingDetector.
+     */
+    private static final long DEFAULT_RELOADING_REFRESH_DELAY_MILLIS = 5000;
+
+    /** How often the periodic trigger asks the reloading controller to check for changes. */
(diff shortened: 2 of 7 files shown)


NicoPiel (dismissed):

mgaffigan:
Sorry @NicoPiel, I had missed a push.  Thanks for the review :)

NicoPiel (approved):
No worries. I love these kinds of speed-ups.

pacmano1 (approved):
Approving.

I'd probably comment on the changed address you used, so nobody "helpfully" fixes it later. The old 9s wasn't DNS. `111.1.1.1` is a valid literal, no lookup. It fails at `bind()` and hits the retry loop in `TcpDispatcher.createServerSocket()`: 10 tries, hardcoded `sleep(1000)` each.

`256.256.256.256` isn't a valid literal, so `getByName` does a DNS lookup instead. Normally a fast NXDOMAIN, so it throws `UnknownHostException` and skips the loop. That's the speedup.

Fine for a test that just checks startup fails with `ConnectorTaskException`. Two things worth capturing: it's invalid on purpose (make it routable and the 9s loop is back), and it leans on the resolver returning NXDOMAIN fast. A slow or hijacking resolver brings the wait back the other way.

mgaffigan:
@pacmano1, agreed on all points. 

I was able to change the timeout on the receiver test (so that one could remain the valid 111.1.1.1), but did not see a clean way to do the same on the dispatcher test. 

Since the test is basically a duplicate of the receiver test, I left it with the DNS error.

jonbartels (approved):
I checked out Mitchs branch. Tests are indeed faster.

I initially had test failures but that was because I had other stuff listening on the common test ports 6667 and 6666. 

I read about the config loading at https://commons.apache.org/proper/commons-configuration/userguide/howto_reloading.html

Non-blocking question - Does naming the constants "DEFAULT_FOO_BAR" imply that it is a default which is configurable, should be configurable, or will be configurable?

Seeing the "testForks" be overridable in gradle is good. If there is some case where more, or less, parallel runners go that helps us tune things.

I see how the TCPReceiver test extends the tcp receiver class and reduces its bind interval during testing.

mgaffigan:
> I initially had test failures but that was because I had other stuff listening on the common test ports 6667 and 6666. 

Yeah. This is an opportunity for future to use auto assigned ports. It required a bunch of changes to the API, so I stopped going down that path in this PR. 

> Non-blocking question - Does naming the constants "DEFAULT_FOO_BAR" imply that it is a default which is configurable, should be configurable, or will be configurable?

Not at Gradle, but each test may decide whether to use a different value.

PR #393, opened 2026-07-29
Add configurable request header size to the HTTP Listener

### LINKED ISSUE #392: [IDEA] Add Request Header Size on HTTP Listener
Add Request Header Size on HTTP Listener as it was done on BridgeLink: https://github.com/Innovar-Healthcare/BridgeLink/issues/129

Image

pacmano1:
Closes #392.

Jetty caps the combined size of all request headers at 8192 bytes and answers anything larger with `431 Request Header Fields Too Large` during header parsing, so the channel never sees the request. This adds a per-connector **Request Header Size** field.

The value is a String and resolves through the template replacer at channel start, like `host`, `port` and `timeout` already do.

## Invalid values fail the connector

Addressing @mgaffigan's review. A value that does not resolve to a positive number now fails the connector at start rather than falling back to 8192:

```java
String requestHeaderSizeValue = replacer.replaceValues(getConnectorProperties().getRequestHeaderSize(), channelId, channelName);
requestHeaderSize = NumberUtils.toInt(requestHeaderSizeValue, 0);

if (requestHeaderSize <= 0) {
    throw new ConnectorTaskException("Invalid request header size: " + requestHeaderSizeValue);
}
```

Falling back was a silent loss of the limit. Someone setting a cap lower than the default, to bound memory use or to stay within what a downstream component expects, would have had the higher default quietly restored.

Passing `0` as the `NumberUtils` default is what makes this one branch instead of two. An unparseable value, an unresolved template, zero and a negative all land on 0 and fail the same way, and the message carries the value as it resolved, so a broken substitution shows up as the literal `${...}` text rather than a number.

Only `requestHeaderSize` changed. The pre-existing `port`, `timeout` and `responseStatusCode` fallbacks in this file have live channels behind them, so tightening those is a separate change.

Addressing @tonygermano's review, the Jetty `HttpConfiguration` is now built in a `protected createHttpConfig(HttpReceiver)` rather than inline in `configureReceiver`. An implementation that overrides `configureReceiver` can build its listener from the current defaults instead of repeating them, and picks up later additions for free. No behaviour change. `TLSHttpConfiguration` in the NovaMap TLS manager plugin extends this class and is the case in point: its non-TLS branch already delegates to `super` and gets the setting, while its TLS branch repeats two of the settings and misses the header size. Raised there as [tls-manager-plugin#44](https://github.com/NovaMap-Health/tls-manager-plugin/issues/44); TLS listeners keep ignoring the setting until that lands.

## Existing channels

Nothing happens to them on upgrade. A channel saved before this feature has no `requestHeaderSize` element, and XStream leaves the field null when it loads one, so `getRequestHeaderSize()` returns 8192 and the channel deploys and runs exactly as it did before. No migration runs and no stored channel is modified.

Open such a channel in the editor and the field shows 8192. That is the getter's answer rather than something read from the channel, so the element is only written the first time someone saves it. Until then an export of that channel still has no element.

The strict check only applies to a value that is actually present, so an absent one can never fail a deploy.

## How to verify

Run on a live engine, HTTP Listener on port 9000, context path `/test`.

| configured value | result |
|---|---|
| `10000` | starts; 9945-byte header returns 200, 9950-byte returns 431 |
| `${headerSize}` with `$gc('headerSize','10000')` in the deploy script | starts; same 9945/9950 boundary, so the resolved value applied rather than the default |
| `0` pushed via the REST API | channel STOPPED, `ConnectorTaskException: Invalid request header size: 0` |
| `${headerSize}` with the deploy script cleared | channel STOPPED, `ConnectorTaskException: Invalid request header size: ${headerSize}` |
| element stripped from the channel XML entirely | starts on 8192, boundary drops accordingly |

The ~55-byte gap between the configured value and the observed boundary is the request line plus `Host`, `User-Agent` and `Accept`.

The last three cannot be reached through the connector panel, which rejects them, so they were pushed through the REST API the way an import or an API-created channel would arrive.

## Tests

12 across three classes: a real Jetty connector for 431 versus 200, a real receiver covering the resolution path including zero, negative, unparseable and an unresolved template, and a serialized-XML round trip proving a channel with no element still gets 8192. Removing the throw fails four tests; changing the default constant fails a third class.

### DIFF
--- a/server/src/main/java/com/mirth/connect/connectors/http/DefaultHttpConfiguration.java
+++ b/server/src/main/java/com/mirth/connect/connectors/http/DefaultHttpConfiguration.java
@@ -47,7 +47,8 @@ public class DefaultHttpConfiguration implements HttpConfiguration {
         org.eclipse.jetty.server.HttpConfiguration httpConfig = new org.eclipse.jetty.server.HttpConfiguration();
         httpConfig.setSendServerVersion(false);
         httpConfig.setSendXPoweredBy(false);
-        
+        httpConfig.setRequestHeaderSize(connector.getRequestHeaderSize());
+
         ServerConnector listener = new ServerConnector(connector.getServer(), new HttpConnectionFactory(httpConfig));
         listener.setHost(connector.getHost());
         listener.setPort(connector.getPort());
--- a/server/src/main/java/com/mirth/connect/connectors/http/DefaultHttpConfiguration.java
+++ b/server/src/main/java/com/mirth/connect/connectors/http/DefaultHttpConfiguration.java
@@ -44,16 +44,24 @@ public class DefaultHttpConfiguration implements HttpConfiguration {
 
     @Override
     public void configureReceiver(HttpReceiver connector) throws Exception {
-        org.eclipse.jetty.server.HttpConfiguration httpConfig = new org.eclipse.jetty.server.HttpConfiguration();
-        httpConfig.setSendServerVersion(false);
-        httpConfig.setSendXPoweredBy(false);
-        httpConfig.setRequestHeaderSize(connector.getRequestHeaderSize());
-
-        ServerConnector listener = new ServerConnector(connector.getServer(), new HttpConnectionFactory(httpConfig));
+        ServerConnector listener = new ServerConnector(connector.getServer(), new HttpConnectionFactory(createHttpConfig(connector)));
         listener.setHost(connector.getHost());
         listener.setPort(connector.getPort());
         listener.setIdleTimeout(connector.getTimeout());
         connector.getServer().addConnector(listener);
+    }
+
+    /**
+     * Builds the Jetty configuration the listener is created with. An implementation that overrides
+     * {@link #configureReceiver(HttpReceiver)} should build its own listener from this rather than
+     * repeating the settings, so that it keeps up with changes to the defaults.
+     */
+    protected org.eclipse.jetty.server.HttpConfiguration createHttpConfig(HttpReceiver connector) {
+        org.eclipse.jetty.server.HttpConfiguration httpConfig = new org.eclipse.jetty.server.HttpConfiguration();
+        httpConfig.setSendServerVersion(false);
+        httpConfig.setSendXPoweredBy(false);
+        httpConfig.setRequestHeaderSize(connector.getRequestHeaderSize());
+        return httpConfig;
     }
 
     @Override
--- a/server/src/main/java/com/mirth/connect/connectors/http/HttpReceiver.java
+++ b/server/src/main/java/com/mirth/connect/connectors/http/HttpReceiver.java
@@ -130,6 +130,7 @@ public class HttpReceiver extends SourceConnector implements BinaryContentTypeRe
     private String host;
     private int port;
     private int timeout;
+    private int requestHeaderSize;
     private String[] binaryMimeTypesArray;
     private Pattern binaryMimeTypesRegex;
     private HttpAuthConnectorPluginProperties authProps;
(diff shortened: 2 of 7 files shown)


kryskool (approved):

server/src/main/java/com/mirth/connect/connectors/http/HttpReceiver.java
```diff
@@ -201,6 +202,16 @@ public void onStart() throws ConnectorTaskException {
...
+        requestHeaderSize = NumberUtils.toInt(replacer.replaceValues(getConnectorProperties().getRequestHeaderSize(), channelId, channelName), HttpReceiverProperties.DEFAULT_REQUEST_HEADER_SIZE);
+
+        /*
+         * Jetty treats a non-positive request header size as no limit at all, so a channel deployed
+         * with one would silently lose the header cap. The connector panel rejects those values, but
+         * a channel imported or pushed through the API never runs that check.
+         */
+        if (requestHeaderSize <= 0) {
+            requestHeaderSize = HttpReceiverProperties.DEFAULT_REQUEST_HEADER_SIZE;
+        }
```
mgaffigan, line 218: Silent ignores are hard to troubleshoot. Can we fail the connector instead?  (Similar to a binding error)

mgaffigan (changes requested):
Non-blocking: Does this really need to be a connector level config? Seems like it could be a much smaller change when coming from mirth.properties.

server/src/main/java/com/mirth/connect/connectors/http/HttpReceiver.java
pacmano1: I don't want the setting on all listeners.   That's the reason it is connector level and not mirth.properties which would be easy of course.   

On the silent ignore, I did go back and forth on that a bit.   It's a coin toss up.  Thoughts on middle ground?  Throw a message to the console, etc?

mgaffigan: That's fine re: making it per connector.  Not something I have a need for, but I understand.

On silent ignore: I retract if it gets three other reviews, but I oppose any soft action here.  If someone is setting a max header size, they may be setting it lower than the default.  A failure to apply such a setting is tantamount to a failure to bind to the correct port or IP.  Writ large:

1. It's likely a DoS vector: If someone tries to set a lower header cap to prevent memory exhaustion, an invalid format quietly restores the higher default cap.
2. It creates quietly broken assumptions: Upstream/downstream logic relying on strict header size bounds (to prevent header smuggling or injection) will silently lose that defense.

It's harder to troubleshoot and likely a security vulnerability.

Note: the same can also be said for use of `NumberUtils.toInt` - but I missed that.

pacmano1: Addressed in latest update.

pacmano1: Done in 82dcd3626. Fails the connector now, and passing `0` as the `NumberUtils` default takes the unparseable case with it. Left `port`, `timeout` and `responseStatusCode` alone since those have live channels behind them.

mgaffigan (dismissed):

server/src/main/java/com/mirth/connect/connectors/http/DefaultHttpConfiguration.java
```diff
@@ -47,7 +47,8 @@ public void configureReceiver(HttpReceiver connector) throws Exception {
         org.eclipse.jetty.server.HttpConfiguration httpConfig = new org.eclipse.jetty.server.HttpConfiguration();
         httpConfig.setSendServerVersion(false);
         httpConfig.setSendXPoweredBy(false);
-        
+        httpConfig.setRequestHeaderSize(connector.getRequestHeaderSize());
```
tonygermano: suggestion: pull out the creation of the httpConfig object to a protected method

Plugins, such as the novamap tls plugin, override the configureReceiver method, so they will not automatically pick up this change. Calling a separate method to create the httpConfig object would allow them to pick up any future changes to the default version of this object without needing to copy/paste the same code every time.

pacmano1: Done in 22f16c053, as `createHttpConfig` rather than `createHttpConfiguration` since this class already implements an `HttpConfiguration` and the Jetty type has to be fully qualified as it is.

Checked `TLSHttpConfiguration` and it extends this class, so the protected method reaches it. Its TLS branch repeats `setSendServerVersion` and `setSendXPoweredBy` and misses the header size, so it can drop to `createHttpConfig(connector)` plus its own `SecureRequestCustomizer`. Its non-TLS branch already delegates to `super` and was picking the setting up anyway. Raised on their side as a follow-up; TLS listeners keep ignoring it until that lands.

Left the host/port/idleTimeout block alone since their listener wraps `SslConnectionFactory` and uses `setConnectors`, so there's no clean shared seam.

mgaffigan (approved):

tonygermano (approved):

jonbartels:
@pacmano1  can you rebase against main plz?

pacmano1:
@jonbartels rebased onto main, CI green. No changes beyond the rebase.

@mgaffigan @tonygermano the force-push staled your approvals. Sorry, re-requesting.

mgaffigan (approved):

jonbartels (approved):

kryskool (approved):

PR #395, opened 2026-07-29
Fix #309: make channelName available in Postprocessor script

### LINKED ISSUE #309: [BUG] channelName variable is not available in Postprocessor script
**Describe the bug**
The `channelName` variable is **not available** in the Postprocessor script. It makes no sense that is not available, since it is available for the Deploy, Undeploy and Preprocessor scripts.

**To Reproduce**
1. Create a channel with a Postprocessor script that makes use of the `channelName` variable.
2. Send a message to the channel and observe that the variable does not exist.

**Expected behavior**
The variable `channelName` should be available in the Postprocessor script.

**Actual behavior**
The variable `channelName` is not available in the Postprocessor script.

**Screenshots**

Image
Image

**Environment:**
- Docker image: [4.5.2-ubuntu](https://hub.docker.com/layers/openintegrationengine/engine/4.5.2-ubuntu)
 - OS: Linux
 - Java Distribution/Version: OpenJDK 17
 - Connect Version: 4.5.2

**Workaround(s)**
Obtain the channel name in other place that where is available and store it in another variable that the Postprocessor script has access to.

**Additional context**
There was already a bug report for Mirth Connect, but it was automatically closed without a fix: https://github.com/nextgenhealthcare/connect/issues/3893

ggiannola:
## Summary

Fixes #309. The `channelName` variable was not available in Postprocessor scripts, even though it is available in Deploy, Undeploy, and Preprocessor scripts.

## Root cause

The Postprocessor scope (`JavaScriptScopeUtil.getPostprocessorScope`) derives `channelName` from `message.getMergedConnectorMessage().getChannelName()`.

However, `Message.getMergedConnectorMessage()` built the merged `ConnectorMessage` setting `channelId`, `messageId`, `serverId`, and `receivedDate` — but never `channelName`. The value therefore resolved to `null` in Postprocessor scripts.

This also explains why the other script contexts were unaffected: they receive the channel name directly (Deploy/Undeploy via `getDeployScope`/`getUndeployScope`, Preprocessor from the connector message) rather than through the merged connector message.

## Fix

In `Message.getMergedConnectorMessage()`, populate `channelName` on the merged connector message:

- from the `Message`'s own `channelName` field, and
- falling back to the source connector message's `channelName`, which is reliably set during processing (see the `ConnectorMessage` construction in `Channel`).

The fallback matters because the `Message` created during processing does not always carry `channelName`, while the source connector message does.

## Tests

Added `MessageTest` covering three cases:

1. The merged message inherits `channelName` from the source connector message.
2. The merged message uses the `Message`-level `channelName` when set.
3. `channelName` stays `null` when unavailable, so the fix does not fabricate a value.

Verified with `./gradlew :donkey:test`: 21 tests pass, 0 failures (`MessageTest` 3, `StatisticsTest` 9, `JdbcDaoTest` 8, `ChannelTest` 1).

### DIFF
--- a/donkey/src/main/java/com/mirth/connect/donkey/model/message/Message.java
+++ b/donkey/src/main/java/com/mirth/connect/donkey/model/message/Message.java
@@ -117,6 +117,7 @@ public class Message implements Serializable {
         if (mergedConnectorMessage == null) {
             mergedConnectorMessage = new ConnectorMessage();
             mergedConnectorMessage.setChannelId(channelId);
+            mergedConnectorMessage.setChannelName(channelName);
             mergedConnectorMessage.setMessageId(messageId);
             mergedConnectorMessage.setServerId(serverId);
             mergedConnectorMessage.setReceivedDate(receivedDate);
@@ -128,6 +129,9 @@ public class Message implements Serializable {
             ConnectorMessage sourceConnectorMessage = connectorMessages.get(0);
 
             if (sourceConnectorMessage != null) {
+                if (mergedConnectorMessage.getChannelName() == null) {
+                    mergedConnectorMessage.setChannelName(sourceConnectorMessage.getChannelName());
+                }
                 mergedConnectorMessage.setRaw(sourceConnectorMessage.getRaw());
                 mergedConnectorMessage.setProcessedRaw(sourceConnectorMessage.getProcessedRaw());
                 sourceMap = sourceConnectorMessage.getSourceMap();
--- a/donkey/src/test/java/com/mirth/connect/donkey/test/MessageTest.java
+++ b/donkey/src/test/java/com/mirth/connect/donkey/test/MessageTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) Mirth Corporation. All rights reserved.
+ * 
+ * http://www.mirthcorp.com
+ * 
+ * The software in this package is published under the terms of the MPL license a copy of which has
+ * been included with this distribution in the LICENSE.txt file.
+ */
+
+package com.mirth.connect.donkey.test;
+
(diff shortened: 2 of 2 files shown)


pacmano1:
@ggiannola, this works in the post processor.

```JS
logger.info(ChannelUtil.getChannelName(channelId))
```

Not that your PR is "wrong", but based on this trivial workaround it seems unnecessary.

jonbartels (approved):
Approved. 

I have a non-blocking comment - Are there other fields missing from the post-processor?

You found that channelName was missing. Are there other common fields from other scripts that are missing that could also be added in this PR?

ggiannola:
> @ggiannola, this works in the post processor.
> 
> ```js
> logger.info(ChannelUtil.getChannelName(channelId))
> ```
> 
> Not that your PR is "wrong", but based on this trivial workaround it seems unnecessary.

Thanks @pacmano1  agreed.

Worth noting the postprocessor scope already tries to supply channelName; it was wired in deliberately back in 2015 (2fd0148dd2, MIRTH-1763) and has silently been null ever since because Message.getMergedConnectorMessage() never populated the field. So this is a latent bug rather than a new feature.

It's also broader than the variable, message.getMergedConnectorMessage().getChannelName() is public userutil API and returns null in every postprocessor, which ChannelUtil doesn't address.

4 lines, no API change. But if you'd rather leave it, happy to close.

kryskool (approved):

mgaffigan (approved):

pacmano1:
@ggiannola , I'll defer to the other maintainers.   I am wishy washy on it, lol.

tonygermano (approved):
Looks good to me. Thanks for the fix and added tests!

Issue #153, opened 2025-07-30
[BUG] When I clone a channel, the dependencies are not carried over to the cloned channel


anibal2j:
**Describe the bug**
I have a set of channels that I usually clone since they're the basis of my configuration for the VA. Each VISN has its own channel and as such when I get a new channel to create I clone from a known last good channel configuration.

The cloning works great except that the channel cloned does not have the dependencies selected as the channel it was cloned from. So, the code in the cloned channel does not really work.

**To Reproduce**
1. Create a channel and make sure you have at least one dependency
2. Clone channel
3. Verify that cloned channel doesn't have the selected dependency from the channel you cloned from

**Expected behavior**
The cloned channel should have the same dependencies already selected as the channel it was cloned from

**Actual behavior**
The cloned channel doesn't have any dependencies selected at all - which makes the channel not work with all the code that's in the source/destination filters/transformers/etc.

**Environment (please complete the following information):**
 - OS: server running on Ubuntu. Client running on Mac OS 15.5 (24F74)
 - Java Distribution/Version 

openjdk version "11.0.28" 2025-07-15
OpenJDK Runtime Environment (build 11.0.28+6-post-Ubuntu-1ubuntu122.04.1)
OpenJDK 64-Bit Server VM (build 11.0.28+6-post-Ubuntu-1ubuntu122.04.1, mixed mode, sharing)

 - Connect Version : 4.4.2

**Workaround(s)**
My workaround now is to go to the channel I'm cloning from, looking at the dependencies, either memorizing them, taking them down somewhere, or screenshoting, then go to the newly created channel that was cloned and manually find the dependencies and select them. When I have a channel that has more than 5 dependencies this becomes pretty error prone and time consuming.

mgaffigan:
Related:
- https://github.com/nextgenhealthcare/connect/issues/4206
- https://github.com/nextgenhealthcare/connect/issues/4073
- https://github.com/nextgenhealthcare/connect/issues/4072

kpalang:
> +1 for this. Cloned channel should also go into the same group as the cloned channel. If not edited after cloning, a cloned channel should be identical to the original except for the name in every way.

Without actually looking at the code, putting the channel in to the same group on server B as it was on server A is a bit trickier than it seems. As I see it there are two obstacles to this:
1. Groups (like channels) are identified by UUID-s. This means the groups need to have identical IDs on both servers. This is absolutely doable, but is not the default behavior.
[...]

PR #396, opened 2026-07-29
Add configurable channel clone dialog

eddcomeau:
## Summary

Implements Issue #153 by adding a configurable channel clone dialog.
image

https://github.com/OpenIntegrationEngine/engine/issues/153

Users can now choose what to copy when cloning a channel, including:

- Channel name, with validation
- Prune settings
- Group membership
- Tags
- Outbound channel dependencies
- Code Template Library assignments
- Shared resource assignments

The clone receives a new GUID and revision zero, preserves the source channel, and opens in the normal channel editor for review before saving.

## Additional fixes

- Default clone name is `<source name>1`.
- Tags use the shared channel-tag association path so they persist correctly and appear in Settings → Tags.
- Inbound dependencies are not copied or modified.
- Existing clone behavior remains available through the new dialog.

## Testing

- Client compilation passes with:

```powershell
.\gradlew.bat :client:compileJava --no-daemon
```

Compiled and tested in the Client/Server on Windows10:
- Dependencies are carried over and can be toggled.
- Prune Settings are carried over and can be toggled.
- Tags are carried over and can be toggled.
- Group Membership are carried over and can be toggled.
- Code Templates are carried over and can be toggled.
- Resources are carried over and can be toggled.

Hard to capture in pictures, but proof that I went in and tested the endpoints at least.
image

### DIFF
--- a/client/src/main/java/com/mirth/connect/client/ui/ChannelCloneDialog.java
+++ b/client/src/main/java/com/mirth/connect/client/ui/ChannelCloneDialog.java
@@ -1,0 +1,286 @@
+/*
+ * Copyright (c) Mirth Corporation. All rights reserved.
+ *
+ * The software in this package is published under the terms of the MPL license.
+ */
+
+package com.mirth.connect.client.ui;
+
+import java.awt.Dimension;
+import java.awt.Window;
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import javax.swing.DefaultComboBoxModel;
+import javax.swing.DefaultListModel;
+import javax.swing.JButton;
+import javax.swing.JCheckBox;
+import javax.swing.JComboBox;
+import javax.swing.JLabel;
+import javax.swing.JList;
+import javax.swing.JOptionPane;
+import javax.swing.JPanel;
+import javax.swing.JScrollPane;
+import javax.swing.JTextField;
+import javax.swing.ListSelectionModel;
+
+import com.mirth.connect.model.Channel;
+import com.mirth.connect.model.ChannelDependency;
+import com.mirth.connect.model.ChannelGroup;
+import com.mirth.connect.model.ChannelTag;
+
+import net.miginfocom.swing.MigLayout;
+import org.apache.commons.lang3.StringUtils;
+
+/** Dialog for selecting which channel associations should be copied during a clone. */
+public class ChannelCloneDialog extends MirthDialog {
+
+    public interface ChannelNameValidator {
+        boolean isValid(String name);
+    }
+
+    public static class Option {
+        private final String id;
+        private final String name;
+
+        public Option(String id, String name) {
+            this.id = id;
+            this.name = name;
+        }
+
+        public String getId() {
+            return id;
+        }
+
+        @Override
+        public String toString() {
+            return name == null ? id : name;
+        }
+    }
+
+    private final List<Option> allTags;
+    private final List<Option> allDependencies;
+    private final Set<String> sourceTagIds;
+    private final String sourceId;
+    private final ChannelNameValidator channelNameValidator;
+    private boolean confirmed;
+
+    private JTextField nameField;
+    private JCheckBox pruneCheckBox;
+    private JCheckBox codeTemplateCheckBox;
+    private JCheckBox resourcesCheckBox;
+    private JComboBox<GroupOption> groupComboBox;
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/mirth/connect/client/ui/ChannelCloneDialog.java
```diff
@@ -0,0 +1,285 @@
...
+        setTitle("Clone channel \"" + channel.getName() + "\"");
+        setPreferredSize(new Dimension(650, 540));
+        setDefaultCloseOperation(DISPOSE_ON_CLOSE);
+        pack();
+        setLocationRelativeTo(owner);
+        setVisible(true);
+    }
+
+    private void initComponents(Channel channel, List<GroupOption> groups, String selectedGroupId, boolean hasCodeTemplates, boolean hasResources) {
+        nameField = new JTextField(channel.getName() + "1");
```
kryskool: The channel length is 40 characters, so it must be truncated to 39 before adding "1".

jonbartels:
> Users can now choose what to copy when cloning a channel

Can you show what the selection dialog looks like? Add it as an edit to the issue description.

client/src/main/java/com/mirth/connect/client/ui/ChannelPanel.java
mgaffigan, line 1: This is already a very large and overcomplicated file. Can we avoid clone logic being added here?

client/src/main/java/com/mirth/connect/client/ui/ChannelSetup.java
```diff
@@ -524,7 +528,7 @@ public void editChannel(Channel channel) {
         channelValidationFailed = false;
         lastModelIndex = -1;
         currentChannel = channel;
-        saveGroupId = null;
+        saveGroupId = groupId;
```
mgaffigan, line 531: More global state? Can we avoid?

mgaffigan (commented):
Thanks for the PR! I'm not blocking if you have three other reviewers.

Over time, we have to simplify the core channel operations - this is a perpetuation of the global state everywhere with logic on the client.  

I'd prefer to see either:
- if clone does not require service calls: a new "clone" method on channel that returns the unsaved object for edits
- if clone requires service calls: a new "clone" service operation that takes a source channel ID and clone options, and returns the new channel

I would caution against overcomplicating a clone dialog: after clone, the user must edit the channel anyway, and all the editors are existing in that interface.

I think this is also likely related to:
- move channel group membership to be a property of channels, not of groups
- move channel code template use to be a property of channels, not of code templates

Those refactorings may substantially simplify the clone process.

client/src/main/java/com/mirth/connect/client/ui/ChannelCloneDialog.java
eddcomeau: Thanks, fixed this up with a commit https://github.com/OpenIntegrationEngine/engine/commit/84e961643160bb4ed13a20c95fc773621ca95ed0

eddcomeau:
> Thanks for the PR! I'm not blocking if you have three other reviewers.
> 
> Over time, we have to simplify the core channel operations - this is a perpetuation of the global state everywhere with logic on the client.
> 
> I'd prefer to see either:
> 
> * if clone does not require service calls: a new "clone" method on channel that returns the unsaved object for edits
> * if clone requires service calls: a new "clone" service operation that takes a source channel ID and clone options, and returns the new channel
> 
> I would caution against overcomplicating a clone dialog: after clone, the user must edit the channel anyway, and all the editors are existing in that interface.
> 
> I think this is also likely related to:
> 
> * move channel group membership to be a property of channels, not of groups
> * move channel code template use to be a property of channels, not of code templates
> 
> Those refactorings may substantially simplify the clone process.

I appreciate this. I think if I do go the route you talk about here I would do it in a new PR rather than try to salvage this branch. I might leave this up in case the team thinks this is a worthwhile addition in the meantime. 

I’ve been thinking about the longer-term roadmap:
Move channel group membership to the channel side of the model.
Move channel code-template-library assignments to the channel side, with library views deriving their channel lists.
Move channel tag associations to the channel side, while retaining shared tag definitions.
Implement cloning as a dedicated service operation that accepts clone options and returns the new channel.

That all sounds daunting, especially thinking about channel export/import and the need to change server/client/CLI portions in conjunction. Selfishly I like MirthSync and I don't want to stray too far on our export structure. But I haven't thought too hard about these, so I will commit to looking into them at least!

eddcomeau:
> > Users can now choose what to copy when cloning a channel
> 
> Can you show what the selection dialog looks like? Add it as an edit to the issue description.

I added it up there thanks

PR #397, opened 2026-07-29
Fix #120: start and stop each server plugin only once

### LINKED ISSUE #120: [BUG] The ServerPlugin.close() function could be called multiple times when the server shuts down.
**Describe the bug**
Not sure how to explain this one. I'm working on a plugin for OIE (a database connection pools manager) and I was struggling with some errors appearing on the mirth.log file on server shutdown. 

If you look at the [initPlugins](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L167) function, you can see that each plugin will be put in a specific map according to their plugin type ([here](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L241), [here](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L248), [here](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L255), etc.), but they are also added to an [ArrayList of server plugins](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L86C52-L86C61).

Now if you look at the [stopPlugins](https://github.com/OpenIntegrationEngine/engine/blob/main/server/src/com/mirth/connect/server/controllers/DefaultExtensionController.java#L387) function that is called when the server shuts down, you'll see that the `stop` function is called for each plugin in the array list, thus, multiple times if you have a class that implements multiple ServerPlugin.

**To Reproduce**
Create a plugin that implements multiple "ServerPlugin" interface. Let's say I have a class that implements the ServicePlugin interface that also implements the ChannelPlugin (cuz the service needs to intercept channel creation, save, deletion, etc.). Then the "stop()" function will be called twice.

**Expected behavior**
The "stop" function should be called only once.

**Actual behavior**
The "stop" function is called according to the number of "ServerPlugin" interfaces the class implements.

**Workaround(s)**
Create a dedicated class for each type of ServerPlugin you want to implement.

**Additional context**
I would be happy to work on this one if the community thinks it would nice to solve this.

ggiannola:
## Summary

Fixes #120. A plugin whose class implements several of the plugin type interfaces was registered once per interface, so it was started and stopped multiple times.

## Root cause

`DefaultExtensionController.initPlugins()` checks each loaded plugin against every plugin type in turn, and each of the eight branches adds the instance to the same `serverPlugins` collection:

```java
if (serverPlugin instanceof ServicePlugin) { ... serverPlugins.add(servicePlugin); }
if (serverPlugin instanceof ChannelPlugin) { ... serverPlugins.add(channelPlugin); }
// ...six more
```

Because that collection was an `ArrayList`, a class implementing both `ServicePlugin` and `ChannelPlugin` ends up in it twice.

Note the issue title mentions `close()`; the methods actually affected are `start()` and `stop()` on `ServerPlugin`.

## Impact beyond start/stop

The duplicate entries affect every consumer of that collection, not just shutdown:

- `startPlugins()` / `stopPlugins()` invoke `start()` and `stop()` once per implemented interface instead of once per plugin. This is the shutdown noise reported in the issue.
- `AlertWorker` builds its `alertActionAcceptors` list by iterating `getServerPlugins()` and filtering on `instanceof AlertActionAcceptor`, so a duplicated plugin is added repeatedly and its alert actions run once per implemented interface.
- `DefaultUsageController` reports plugin usage from the same collection, so counts are inflated for multi-interface plugins.

## Fix

`serverPlugins` is now a `Set`, so a plugin registered for several interfaces is held once and the de-duplication is the natural behaviour of the collection. The eight registration calls are unchanged.

`LinkedHashSet` rather than `HashSet`: `initPlugins()` deliberately loads plugins in descending plugin weight order, and preserving that order means they are started and stopped in the intended sequence.

`getServerPlugins()` still returns `List<ServerPlugin>`, so the `ExtensionController` signature is unchanged for extensions. It now returns a copy rather than the live internal collection.

Net change is 10 insertions and 2 deletions in one file.

## On testing

The earlier revision of this PR routed the registrations through a de-duplicating helper and unit tested it. Following review that has been dropped in favour of the `Set`, and the tests went with it: they exercised the helper directly, and the invariant is now enforced by the collection type rather than by logic of our own.

Testing this through `initPlugins()` instead is not really reachable from a unit test, since it requires plugin metadata loading plus database access via `setPluginProperties`. If you would like coverage here, I am happy to look at an integration-level test.

Verified with `./gradlew :server:test` (613 tests, 0 failures) and `./gradlew :donkey:test`.

## Note on the failing check

The `FileReceiverTest > testPoll1` failure on the previous commit is unrelated to this change, which touches only plugin registration. The assertion at `FileReceiverTest.java:133` counts files picked up by a poll, and it passes locally both in the full suite and across three consecutive isolated runs. Given the recent "Allow parallel tests" change, this looks like a timing-sensitive flake rather than a real regression.

## How to verify

Build a plugin whose class implements two `ServerPlugin` sub-interfaces, for example `ServicePlugin` and `ChannelPlugin`, and log a line in `stop()`. Before this change the line appears twice on server shutdown; after it, once.

### DIFF
--- a/server/src/main/java/com/mirth/connect/server/controllers/DefaultExtensionController.java
+++ b/server/src/main/java/com/mirth/connect/server/controllers/DefaultExtensionController.java
@@ -84,7 +84,14 @@ public class DefaultExtensionController extends ExtensionController {
 
     // these are plugins for specific extension points, keyed by plugin name
     // (not path)
-    private List<ServerPlugin> serverPlugins = new ArrayList<ServerPlugin>();
+    /*
+     * A plugin class may implement several of the plugin type interfaces, in which case it is
+     * registered once for each interface it implements. A Set holds a single entry per plugin,
+     * so start() and stop() are invoked once per plugin rather than once per interface.
+     * LinkedHashSet because initPlugins loads plugins in a deliberate order (by plugin weight)
+     * and that order is preserved when they are started and stopped.
+     */
+    private Set<ServerPlugin> serverPlugins = new LinkedHashSet<ServerPlugin>();
     private Map<String, ServicePlugin> servicePlugins = new LinkedHashMap<String, ServicePlugin>();
     private Map<String, ChannelPlugin> channelPlugins = new LinkedHashMap<String, ChannelPlugin>();
     private Map<String, CodeTemplateServerPlugin> codeTemplateServerPlugins = new LinkedHashMap<String, CodeTemplateServerPlugin>();
@@ -240,42 +247,42 @@ public class DefaultExtensionController extends ExtensionController {
                          */
                         servicePlugin.init(currentProperties);
                         servicePlugins.put(servicePlugin.getPluginPointName(), servicePlugin);
-                        addServerPlugin(servicePlugin);
+                        serverPlugins.add(servicePlugin);
                         logger.debug("sucessfully loaded server plugin: " + serverPlugin.getPluginPointName());
                     }
 
                     if (serverPlugin instanceof ChannelPlugin) {
                         ChannelPlugin channelPlugin = (ChannelPlugin) serverPlugin;
                         channelPlugins.put(channelPlugin.getPluginPointName(), channelPlugin);
-                        addServerPlugin(channelPlugin);
+                        serverPlugins.add(channelPlugin);
                         logger.debug("sucessfully loaded server channel plugin: " + serverPlugin.getPluginPointName());
                     }
 
                     if (serverPlugin instanceof CodeTemplateServerPlugin) {
                         CodeTemplateServerPlugin codeTemplateServerPlugin = (CodeTemplateServerPlugin) serverPlugin;
                         codeTemplateServerPlugins.put(codeTemplateServerPlugin.getPluginPointName(), codeTemplateServerPlugin);
-                        addServerPlugin(codeTemplateServerPlugin);
+                        serverPlugins.add(codeTemplateServerPlugin);
                         logger.debug("sucessfully loaded server code template plugin: " + serverPlugin.getPluginPointName());
                     }
 
                     if (serverPlugin instanceof DataTypeServerPlugin) {
                         DataTypeServerPlugin dataTypePlugin = (DataTypeServerPlugin) serverPlugin;
                         dataTypePlugins.put(dataTypePlugin.getPluginPointName(), dataTypePlugin);
-                        addServerPlugin(dataTypePlugin);
+                        serverPlugins.add(dataTypePlugin);
                         logger.debug("sucessfully loaded server data type plugin: " + serverPlugin.getPluginPointName());
                     }
 
                     if (serverPlugin instanceof ResourcePlugin) {
                         ResourcePlugin resourcePlugin = (ResourcePlugin) serverPlugin;
                         resourcePlugins.put(resourcePlugin.getPluginPointName(), resourcePlugin);
-                        addServerPlugin(resourcePlugin);
+                        serverPlugins.add(resourcePlugin);
                         logger.debug("Successfully loaded resource plugin: " + resourcePlugin.getPluginPointName());
                     }
 
                     if (serverPlugin instanceof TransmissionModeProvider) {
                         TransmissionModeProvider transmissionModeProvider = (TransmissionModeProvider) serverPlugin;
                         transmissionModeProviders.put(transmissionModeProvider.getPluginPointName(), transmissionModeProvider);
-                        addServerPlugin(transmissionModeProvider);
+                        serverPlugins.add(transmissionModeProvider);
                         logger.debug("Successfully loaded transmission mode provider plugin: " + transmissionModeProvider.getPluginPointName());
                     }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/mirth/connect/server/controllers/DefaultExtensionController.java
```diff
@@ -309,6 +309,32 @@ public void initPlugins() {
...
+     * {@link ServerPlugin#start()} and {@link ServerPlugin#stop()} once per implemented interface
+     * rather than once per plugin.
+     * <p>
+     * Instances are compared by identity on purpose. Two distinct plugin instances must both be
+     * registered even if the plugin class considers them equal.
+     * <p>
+     * Package private so it can be exercised directly by unit tests without going through the full
+     * extension loading process.
+     */
+    void addServerPlugin(ServerPlugin serverPlugin) {
```
jonbartels: What if serverPlugins was a Set instead of a List? That way if `add` was called multiple times there would still only be one instance in the Set. Then you do not need this method to check the list. 

It would use the natural behavior of Set. I think you would want a HashSet for the constant-time behavior, but this is likely to be a fairly small list so performance is not a major concern.

server/src/test/java/com/mirth/connect/server/controllers/DefaultExtensionControllerTest.java
```diff
@@ -72,4 +76,98 @@ public void cleanupTestFolder() {
...
+        extensionController.addServerPlugin(plugin);
+        extensionController.addServerPlugin(plugin);
+
+        List<ServerPlugin> registeredPlugins = extensionController.getServerPlugins();
+        assertEquals(1, registeredPlugins.size());
+        assertSame(plugin, registeredPlugins.get(0));
+    }
+
+    @Test
+    public void testStopPluginsStopsPluginRegisteredForMultipleTypesOnce() {
```
jonbartels: Instead of calling `addServerPlugin` is there a way to more naturally test the behavior?

I am not sure. It may be very ugly to set up the test preconditions in a way that allows `initPlugins` to run and add them to the set

jonbartels (changes requested):
I left some comments in the code with a simpler strategy for the code, though that strategy is not as easily tested.

I also left some related comments on Issue #120  to better understand the conditions. 

I would be inclined to approve this PR if it implemented Set instead of an adder function that required looping over a List. TY for the PR and your first OIE commit!

ggiannola:
Thanks @jonbartels, good call on the Set, that's clearly the right shape. Pushed as `69962aef0`. The helper is gone, the eight registration calls are back to plain `serverPlugins.add(...)`, and the whole change is now +10/-2 in one file.

Two small deviations worth flagging:

- I used `LinkedHashSet` rather than `HashSet`. `initPlugins()` loads plugins in descending weight order and `startPlugins()`/`stopPlugins()` iterate that collection, so `HashSet` would randomize the start/stop sequence. `LinkedHashSet` keeps the Set semantics you asked for and preserves the intended order.
- `getServerPlugins()` still returns `List<ServerPlugin>` since it is declared abstract on `ExtensionController` and changing it would break extensions. It now returns a copy rather than the live collection.

On testing: you were right that it was awkward. The three tests exercised the helper directly, so they went with it. I think that is the better trade: with a Set the invariant is enforced by the collection type rather than by logic of ours, so there is less to regress. Going through `initPlugins()` is not really reachable from a unit test (it needs plugin metadata loading plus DB access via `setPluginProperties`), but happy to look at integration-level coverage if you would like some.

Re your question on #120 about making `stop()` tolerant of multiple calls — that would fix the shutdown errors, but the duplicates also reach `AlertWorker` (duplicate `AlertActionAcceptor`s, so alert actions fire once per implemented interface) and `DefaultUsageController` (inflated plugin usage counts), which plugin-side idempotency would not help with.

One note: the `FileReceiverTest > testPoll1` failure on the previous commit looks like a flake — it counts files picked up by a poll, and it passes locally both in the full suite and across three consecutive isolated runs. The new commit's build is sitting at `action_required`, so it will need a maintainer to approve the workflow run.

server/src/main/java/com/mirth/connect/server/controllers/DefaultExtensionController.java
ggiannola: Agreed, done in `69962aef0`.

Used `LinkedHashSet` rather than `HashSet` , `initPlugins()` loads by descending plugin weight and `startPlugins()`/`stopPlugins()` iterate this collection, so the ordering is meaningful. Also kept `getServerPlugins()` returning a `List` (as a copy), since it is abstract on `ExtensionController` and changing the signature would break extensions.

server/src/test/java/com/mirth/connect/server/controllers/DefaultExtensionControllerTest.java
ggiannola: You were right that it was awkward. With the Set the helper is gone and these tests went with it, the invariant is now enforced by the collection type rather than by logic of ours.

Setting up `initPlugins()` really is ugly: it needs plugin metadata loading plus DB access via `setPluginProperties`, so it is out of reach for a unit test. Happy to look at integration-level coverage if you think it is worth having.

kryskool (approved):

jonbartels (approved):

mgaffigan (approved):

NicoPiel (approved):

Issue #404, opened 2026-08-09
Investigate layered/drop-in configuration for log4j2.properties


pacmano1:
Split out of #50, which is getting mirth.properties.d support for the main config. Logging was deliberately left out of that PR.

log4j2's log4j2.configurationFile accepts a comma-separated list of files and merges them into a CompositeConfiguration, which might make this nearly free. Two things need checking before trusting it: the merge machinery is documented around XML/JSON/YAML and its behavior with properties-format configs (what we ship) needs empirical verification on our log4j version, and the launcher hardcodes ./conf/log4j2.properties in Log4jBootstrap/Log4jMigrations.

Issue #220, opened 2025-12-08
Standardize artifact naming to align with project branding


pacmano1:
To improve consistency and clarity across builds and distributions, update artifact (JAR) names to reflect Open Integration Engine branding rather than legacy naming. This will reduce downstream confusion in deployments, logging, and dependency management as the project continues to evolve independently.

Scope includes:

- Build outputs
- Distribution artifacts
- Any packaging references that surface artifact names externally

No functional changes intended.

Issue #165, opened 2025-08-18
Alter the build pipeline so that build artifacts are branded as "openintegrationengine" or "oie" instead of "mirth"


jonbartels:
To demonstrate our independence from Mirth the build artifacts that are currently generated as `mirth-*.jar` or `mc*` should be generated with OpenIntegrationEngine or OIE branding instead. 

This will correctly represent that these artifacts are from OIE builds and not Mirth builds.

Changing Java package names is out of scope for this ticket.

Some scoping questions:
- do we do this in `ant` or wait for `gradle`? See PR #100 and #54 
- what regressions or unexpected breakage do we need to think about?

sbraconnier:
I personally would wait for Gradle. And would probably also wait until the project is properly split into 1 module per artifact. When the Gradle PR will be merged, it will be easier to do. For example, the source code that produce the `mirth-client-core` artifact should be in its own module, not split when building the project (IMHO).

PR #405, opened 2026-08-09
Load mirth.properties.d drop-in configuration overrides on startup

### LINKED ISSUE #50: [IDEA] Allow multiple MC configuration files t be read on startup
I hate having to re-build mirth.properties and others for each upgrade because there are new things in it. I would love to have a single default config file that ships with every version, but my actual configuration ends up being in a file that doesn't get overwritten and doesn't need to be manually merged-into the new one.

If the product would ship with a default properties file containing certain things (like defaults!) and then you could overlay others, that would be very beneficial.

So instead of a single file to load, it will load a few things, in order.

Always load oie.properties first. Then, load mirth.properties as a really convenient migration strategy for existing users, and overwrite any previous values with stuff from mirth.properties. Then, read local.properties and overwrite again.

Anyone who wants to can use mirth.properties forever. But the product can change it's default default filename to something that doesn't say "mirth" anymore.

If you want to go crazy, you can read files.to.load from oie.properties, then append anything that any of the files.to.load also requests, and so on. So mirth.properties or local.properties can set files.to.load=database.properties and then OIE also loads database.properties even though it wasn't known about in advance.

You could do the same for logging configuration, but I'm not sure how well that would work with the current log4j system. It would be enough to layer primary configuration files (i.e. what is currently mirth.properties).

pacmano1:
Closes #50.

Files named `*.properties` under `conf/mirth.properties.d` are loaded on top of `mirth.properties` at startup, in lexical filename order, later files winning. The server never writes to the directory, so local configuration survives upgrades without hand-merging. This mirrors the existing vmoptions convention (`base_includes.vmoptions` -> `custom.vmoptions`).

All six `mirth.properties` readers apply the overlay: the launcher, the main server, the configuration controller, the WebStart servlet, extension statuses, and the Rhino optimization-level lookup in `JavaScriptScopeUtil`. The configuration controller keeps a separate file-backed config for version migration and password re-encryption, so those writes never bake drop-in values into `mirth.properties`.

**Scope notes:**

- Properties the server encrypts and writes back (`database.password`, keystore passwords) should stay in `mirth.properties`; a plaintext override in a drop-in keeps taking precedence and is never encrypted. Documented in the shipped mirth.properties.example.
- `migrateKeystore()` still reads `mirth.properties` directly; it's a one-shot legacy migration (pre-2.2) and was left alone.
- Layering the log4j2 configuration is split out to #404.
- Renaming the shipped defaults file away from `mirth.properties` (also floated in #50) belongs to the existing branding effort (#165, #220) rather than this PR.

**To verify:** create `conf/mirth.properties.d/10-test.properties` containing `http.port = 8199`, start the server via `oieserver`, confirm the startup log reports the web server on port 8199 and that `conf/mirth.properties` is unchanged afterward.


pacmano1:
Design notes, so the reasoning doesn't live only in my head:

- A drop-in directory over a special sibling file (`custom.properties`, `local.properties`): matches the `conf.d`/`sysctl.d` convention, scales past one file, and the name derives from the parent file. The vmoptions chain (`base_includes` -> `custom.vmoptions`) is the in-repo precedent for layered config.
- The sample ships as `mirth.properties.example` so the loader ignores it; copy it to a `*.properties` name to activate.
- The server does not modify `mirth.properties` in normal operation (verified across a boot: identical content). First boot only trims trailing whitespace on empty values, appends `;upgrade=true` to the Derby URL, and generates keystore passwords. So the directory exists for upgrade hygiene: the installer preserves `mirth.properties`, and tar.gz upgrades can replace it wholesale without hand-merging.
- Keystore gotcha, verified live on a fresh install: `keystore.storepass`/`keystore.keypass` auto-generate only when BOTH are still the shipped default. A drop-in overriding just one suppresses generation and the other stays at its published default. Pre-existing guard; the sample file documents it.
- Drop-ins are parsed with java.util.Properties in every reader so the launcher-side and server-side worlds see identical semantics. `${...}` references would interpolate only on the server side, so the sample avoids suggesting them.
- log4j2 layering split to #404. Renaming the shipped defaults file (also floated in #50) belongs to the branding effort (#165/#220).

mgaffigan (dismissed):
Can we collapse this to a global store?  Each consumer should not have to construct this themselves, nor should we be reading it so many times during boot and operation without reason.

Also also: if we're doing this, can we sneak in or at least leave room for configuration via environment variables taking precedence over everything?

pacmano1:
Both good.

Store: agreed. `DefaultConfigurationController` already loads `mirth.properties` once, so I'll point `Mirth`, `WebStartServlet`, and `JavaScriptScopeUtil` at that copy instead of each reopening the file (`WebStartServlet` re-reads on every JNLP request). `MirthLauncher` and `ExtensionStatuses` run before the server classpath exists, so those two keep their own read. I'll do that here.

Env: reading this as the Docker path. `configure-from-env` writes env into `mirth.properties`, so with these drop-ins on top a drop-in now overrides an env value, backwards from "env wins." Plan: have `configure-from-env` write to a last-sorted drop-in (`99-env.properties`) instead, so env wins. Docker follow-up, not this PR, so this one just leaves room. Shout if you meant something broader.

pacmano1:
**Consolidated config loading.** `Mirth` and `WebStartServlet` no longer each load and merge `mirth.properties` + the drop-ins. They read one shared, already-merged copy from `ConfigurationController.getPropertiesConfiguration()`. `saveMirthConfig()` still writes only the base file, so drop-in values never get baked into `mirth.properties` when the server re-encrypts a password or generates keystore passwords. `JavaScriptScopeUtil` stays self-contained on purpose: its static initializer runs in unit tests with no server, so it can't call the controller.

**A bad drop-in file now fails closed.** Two cases:
- **Bad file** (unreadable, or a malformed `\uXXXX` escape): before, `overlay()` threw during init, the throw was swallowed, and the server came up on the base config with the drop-ins dropped. That's fail-open: if a drop-in was hardening a setting (restricting TLS, forcing HTTPS), the server reverts to the weaker base value with no signal. Now the controller catches it and the server refuses to start with a clear message. The bootstrap readers (`MirthLauncher`, `ExtensionStatuses`) stay lenient and skip a bad file with a warning, matching how they already treat `mirth.properties`; they only read low-stakes values, and the authoritative config is gated on the server path.
- **Bad value in a good file** (`http.port = banana`): not new, and not something the loader can judge. It behaves the same as that value in `mirth.properties` today, so it's left as-is. Real config validation is a separate change.

Also documented that `getPropertiesConfiguration()` hands out the controller's live config and consumers must treat it as read-only. Enforcing that with an immutable view changes the type through the web server and TLS path, so it's documented here for now and I'll file a follow-up to enforce it.

**Verified:** full `clean build`, 683 tests. Booted via `oieserver`: a valid drop-in (`http.port` override) boots clean and binds the overridden port. A malformed drop-in and an unreadable (`chmod 000`) drop-in each make the engine log `Refusing to start` and bind no ports.

pacmano1:
Heads up on CI: the `build` check failed once on this commit, but it was a flaky test (`FileReceiverTest.testPoll1`, a file-count assertion unrelated to this change). It passed on re-run and doesn't reproduce locally on macOS or on a native-Linux run of the same commit, so all checks are green now. Root cause of the flake (server tests share one JVM and `FileReceiverTest` injects a mock `ControllerFactory` into global static state) filed as #414.

pacmano1:
@mgaffigan, this has been waiting on you since August.

The global store is done, in `0a23598f0` on 8/16. `ConfigurationController.getPropertiesConfiguration()` returns mirth.properties with the drop-ins already applied, and `Mirth` and `WebStartServlet` read that one copy instead of each loading and merging the file themselves. The same commit makes a bad drop-in fail closed. Before it, `overlay()` threw during controller init, the exception got swallowed, and the server came up on the base config with the drop-ins gone, which silently undoes whatever a drop-in was hardening.

Environment variables are not in this PR. I went with the "leave room" option. Everything reads through `getPropertiesConfiguration()` now, so that is the one place a precedence layer would go.

I'd rather do that separately, and there's a reason beyond scope. `configure-from-env` seds env vars straight into `conf/mirth.properties`, so it destructively rewrites the file this PR exists to stop people hand-editing, and it only covers a fixed list of properties. Pointing it at `conf/mirth.properties.d/` instead fixes both and needs no Java changes. Say the word and I'll file it.

This is merged up to current main and green on all seven CI configurations. Can you look again, or clear the changes-requested if you're happy?

mgaffigan:
@pacmano1, I've dismissed my review, but do not have time right now to review fully. I certainly don't want to block progress.

I have no problem with the env variables being added in a separate PR.

My concerns are:
1. Are we reading properties consistently throughout the app (+DRY)
2. Do we take any new assumptions on CWD (historical source of fragility for tests and windows service startup)
3. How do writes to the properties file land in the correct place (and what is the correct place)
4. How does this impact upgrades
5. Is the new directory appropriately secured to limit local EOP
6. Does the installer handle the new directory appropriately

PR #406, opened 2026-08-10
Prevent XML xxe on Strict HL7 parsing

mgaffigan:
Avoids any possibility of XXE by not accepting XML input on a channel to receive HL7.  Turns the message into an error - just as it would be for any other invalid format.

```text
ER7Serializer error
ERROR MESSAGE: Error converting ER7 to XML
ca.uhn.hl7v2.parser.EncodingNotSupportedException: Determine encoding for message. The following is the first 50 chars of the message for reference, although this may not be where the issue is: <?xml version="1.0"?>
...
```

Breaking change for anyone relying on being able to send XML to a channel configured to receive HL7.  Opt-out of the new behavior by toggling "Allow XML" in the datatype properties.

image

Thanks to Samuel Paschuan for reporting the issue.

### DIFF
--- a/server/src/main/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7Serializer.java
+++ b/server/src/main/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7Serializer.java
@@ -199,7 +199,7 @@ public class ER7Serializer implements IMessageSerializer {
                 Message message = null;
                 source = source.trim();
 
-                if (source.length() > 0 && source.charAt(0) == '<') {
+                if (serializationProperties.isAllowXml() && source.length() > 0 && source.charAt(0) == '<') {
                     if (serializationProperties.isUseStrictValidation()) {
                         // If the message is XML and strict validation is needed, we'll need to create a message to be encoded.
                         message = serializationXmlParser.parse(source);
--- a/server/src/test/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7SerializerTest.java
+++ b/server/src/test/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7SerializerTest.java
@@ -1,5 +1,6 @@
 package com.mirth.connect.plugins.datatypes.hl7v2;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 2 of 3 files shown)


server/src/test/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7SerializerTest.java
```diff
@@ -53,4 +65,27 @@ public void testValidFromXMLWithExternalDTD() throws Exception {
...
 	}
+
+	@Test
+	public void testToXmlWithStrictParserRejectsXmlInputByDefault() throws Exception {
+		String xmlDisguisedAsHl7 = "<foo><bar>notreallyhl7</bar></foo>";
+
+		boolean exceptionThrown = false;
+		try {
+			strictSerializer.toXML(xmlDisguisedAsHl7);
+		} catch (MessageSerializerException e) {
```
jonbartels, line 76: Should this specifically inspect the exception message and cause? That way it validates it threw because of the XML parsing and not because of some other serialization fault.

What is the output of this test before and after the other code changes?

server/src/test/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7SerializerTest.java
```diff
@@ -53,4 +65,27 @@ public void testValidFromXMLWithExternalDTD() throws Exception {
...
+			exceptionThrown = true;
+		}
+
+		assertTrue(exceptionThrown);
+	}
+
+	@Test
+	public void testToXmlWithStrictParserAllowsXmlInputWhenOptedIn() throws Exception {
+		String xmlDisguisedAsHl7 = "<foo><bar>notreallyhl7</bar></foo>";
```
jonbartels, line 85: Refactor this to a constant since its used in multiple tests

server/src/main/java/com/mirth/connect/plugins/datatypes/hl7v2/ER7Serializer.java
```diff
@@ -199,7 +199,7 @@ public String toXML(String source) throws MessageSerializerException {
                 Message message = null;
                 source = source.trim();
 
-                if (source.length() > 0 && source.charAt(0) == '<') {
```
jonbartels, line 202: Non-blocking - How can this be documented in the release notes?

jonbartels (changes requested):

jonbartels (approved):

Issue #307, opened 2026-05-19
[SECURITY] mirth.properties and keystore.jks are readable to other users on server


mgaffigan:
**Describe the security issue**
If a server/container is not dedicated to the engine, attackers may be able to read mirth.properties and keystore.jks, which can allow escalation of privilege or loss of nonrepudiation through misuse of database and ssl secrets.

**Suggested remediation**
Have the installer and server application set 600-equivalent privileges on the mirth.properties and keystore.jks files.

**Workaround**
Posix: 
```bash
chmod 600 mirth.properties
chmod 600 keystore.jks
```

Windows:
Edit NTFS permissions on the files to:
1. Remove read access to `BUILTIN\Users`
2. Add Full Control to `NT Service\Open Integration Engine`

**Scoring**
CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N **5.5**

**Thanks to**
Guus Verbeek and DIVD

PR #422, opened 2026-08-26
Secure keystores with filesystem acl, use pkcs12

mgaffigan:
Related to #307 

Sets default access controls for key material created on first boot to only be accessible to the owner (600-equivalent):

Posix:
```text
# ls -al server/setup/appdata/keystore.jks
-rw-------@ 1 user  staff  3973 Aug 26 17:23 server/setup/appdata/keystore.jks
```

Windows:
```text
PS C:\drop\setup> icacls .\appdata\keystore.jks
.\appdata\keystore.jks COMPUTER\mirthuser:(F)

Successfully processed 1 files; Failed processing 0 files
PS C:\drop\setup>
```

Removes the 81uWxplDtB password in favor of a password-less PKCS12 (pfx) file instead of the deprecated jks format for new installs. Upgrades retain their prior password.

### DIFF
--- a/server/basedir-includes/configure-from-env
+++ b/server/basedir-includes/configure-from-env
@@ -16,10 +16,18 @@ if [ $custom_extension_count != 0 ]; then
 	done
 fi
 
-# set storepass and keypass to 'changeme' so they aren't overwritten later
-KEYSTORE_PASS=changeme
-sed -i "s/^keystore\.storepass\s*=\s*.*\$/keystore.storepass = ${KEYSTORE_PASS//\//\\/}/" "$APP_DIR/conf/mirth.properties"
-sed -i "s/^keystore\.keypass\s*=\s*.*\$/keystore.keypass = ${KEYSTORE_PASS//\//\\/}/" "$APP_DIR/conf/mirth.properties"
+# The keystore is created with no passphrase and protected by file permissions instead, so there is
+# no longer a password to pin here. Keystores created by earlier versions are JCEKS and were pinned
+# to 'changeme', so keep those settings when one is already present in appdata.
+KEYSTORE_FILE="$APP_DIR/appdata/keystore.pfx"
+LEGACY_KEYSTORE_FILE="$APP_DIR/appdata/keystore.jks"
+if [ -f "$LEGACY_KEYSTORE_FILE" ] || ! [ -z "${KEYSTORE_DOWNLOAD+x}" ]; then
+	echo "Found an existing keystore at ${LEGACY_KEYSTORE_FILE}, using jks keystore defaults."
+	sed -i 's|^keystore\.path\s*=\s*.*$|keystore.path = ${dir.appdata}/keystore.jks|' "$APP_DIR/conf/mirth.properties"
+	sed -i "s/^keystore\.storepass\s*=\s*.*\$/keystore.storepass = changeme/" "$APP_DIR/conf/mirth.properties"
+	sed -i "s/^keystore\.keypass\s*=\s*.*\$/keystore.keypass = changeme/" "$APP_DIR/conf/mirth.properties"
+	sed -i "s/^keystore\.type\s*=\s*.*\$/keystore.type = JCEKS/" "$APP_DIR/conf/mirth.properties"
+fi
 
 # merge the environment variables into /opt/engine/conf/mirth.properties
 # db type
@@ -230,9 +238,9 @@ fi
 if ! [ -z "${KEYSTORE_DOWNLOAD+x}" ]; then
 	echo "Downloading keystore at ${KEYSTORE_DOWNLOAD}"
 	if ! [ -z "${ALLOW_INSECURE}" ] && [ "${ALLOW_INSECURE}" == "true" ]; then
-		curl -ksSLf "${KEYSTORE_DOWNLOAD}" -o "$APP_DIR/appdata/keystore.jks" || echo "problem with keystore download"
+		curl -ksSLf "${KEYSTORE_DOWNLOAD}" -o "$LEGACY_KEYSTORE_FILE" || echo "problem with keystore download"
 	else
-		curl -sSLf "${KEYSTORE_DOWNLOAD}" -o "$APP_DIR/appdata/keystore.jks" || echo "problem with keystore download"
+		curl -sSLf "${KEYSTORE_DOWNLOAD}" -o "$LEGACY_KEYSTORE_FILE" || echo "problem with keystore download"
 	fi
 fi
 
--- a/server/conf/mirth.properties
+++ b/server/conf/mirth.properties
@@ -25,10 +25,13 @@ password.reuselimit = 0
 version = 4.6.0
 
 # keystore
-keystore.path = ${dir.appdata}/keystore.jks
-keystore.storepass = 81uWxplDtB
-keystore.keypass = 81uWxplDtB
-keystore.type = JCEKS
+# Created if missing with no passphrase and secured by chmod 600.
+keystore.path = ${dir.appdata}/keystore.pfx
+keystore.storepass =
+keystore.keypass =
+# Set to JCEKS for compatibility with older versions of the engine.
+# a PKCS12 is a standard pfx file.  JCEKS is a the older java-specific jks format.
+keystore.type = PKCS12
 
 # server
 http.contextpath = /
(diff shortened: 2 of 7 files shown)


server/basedir-includes/configure-from-env
```diff
@@ -16,10 +16,16 @@ if [ $custom_extension_count != 0 ]; then
...
 
-# set storepass and keypass to 'changeme' so they aren't overwritten later
-KEYSTORE_PASS=changeme
-sed -i "s/^keystore\.storepass\s*=\s*.*\$/keystore.storepass = ${KEYSTORE_PASS//\//\\/}/" "$APP_DIR/conf/mirth.properties"
-sed -i "s/^keystore\.keypass\s*=\s*.*\$/keystore.keypass = ${KEYSTORE_PASS//\//\\/}/" "$APP_DIR/conf/mirth.properties"
+# The keystore is created with no passphrase and protected by file permissions instead, so there is
+# no longer a password to pin here. Keystores created by earlier versions are JCEKS and were pinned
+# to 'changeme', so keep those settings when one is already present in appdata.
+KEYSTORE_FILE="$APP_DIR/appdata/keystore.jks"
+if [ -f "$KEYSTORE_FILE" ] && [ "$(head -c 4 "$KEYSTORE_FILE" | od -An -tx1 | tr -d '[:space:]')" = "cececece" ]; then
```
jonbartels: Explan what this is checking and how it knows its a jceks?

server/src/main/java/com/mirth/connect/server/util/FilePermissionUtil.java
```diff
@@ -0,0 +1,81 @@
...
+import java.nio.file.attribute.AclFileAttributeView;
+import java.nio.file.attribute.PosixFileAttributeView;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.nio.file.attribute.UserPrincipal;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Set;
+
+public class FilePermissionUtil {
```
jonbartels, line 14: Non blocking comment - I checked if file utils does this for us. It does not. I dont live introducing a util class but its appropriate here

server/src/test/resources/mirth.properties
```diff
@@ -25,10 +25,13 @@ password.reuselimit = 0
 version = 4.6.0
 
 # keystore
+# Created if missing with no passphrase and secured by chmod 600.
 keystore.path = ${dir.appdata}/keystore.jks
-keystore.storepass = 81uWxplDtB
-keystore.keypass = 81uWxplDtB
-keystore.type = JCEKS
+keystore.storepass =
```
jonbartels, line 29: Per Tonys comments from chat - add a comment in the props file explaining how this is generated if its blank.

jonbartels (commented):
This is making my brain itch. I feel like there could be some cyclic problem, maybe with containers? where the keystore is generated and a PW set, then the keystore is retained as a file (like if appdata is a persistent volume) but the generated keystore pass isn't retained because conf isnt in a persistent volume and the env var isn't set. 

Instinctively - I'd document this, but theres a massive curse of knowledge problem because I am intimate with my keystores but a more casual, first-time user won't know how to process the big error

server/src/main/java/com/mirth/connect/server/util/FilePermissionUtil.java
jonbartels: Maybe `touch` to create the file? But then you still gotta mess with perms. https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html#touch(java.io.File)

mgaffigan: I don't see an alternative here for any major portion.

server/src/test/resources/mirth.properties
mgaffigan: I did on line 28.  With pkcs12, there is no need to have a password, so the blank is not a sentinel value - there just is no password.

server/basedir-includes/configure-from-env
mgaffigan: This is a dirty hack suggested to me by AI.  It's checking if the first 4 bytes of the file are cececece - which is the magic signature for a JCEKS keystore (the legacy format that would require the "changeme" default password).

TL;DR: backcompat hack.

server/src/test/resources/mirth.properties
```diff
@@ -25,10 +25,13 @@ password.reuselimit = 0
 version = 4.6.0
 
 # keystore
+# Created if missing with no passphrase and secured by chmod 600.
 keystore.path = ${dir.appdata}/keystore.jks
```
tonygermano: Should the file extension change to reflect the new file type? I don't know when I'll get time to test this myself, but does the channel level encryption still work when the keystore type is PKCS12?

The second use of this keystore is to store the symmetric key for channel encryption.

server/src/main/java/com/mirth/connect/server/util/FilePermissionUtil.java
```diff
@@ -0,0 +1,81 @@
...
+        if (aclView != null) {
+            UserPrincipal owner = aclView.getOwner();
+            // @formatter:off
+            AclEntry entry = AclEntry.newBuilder()
+                    .setType(AclEntryType.ALLOW)
+                    .setPrincipal(owner)
+                    .setPermissions(EnumSet.allOf(AclEntryPermission.class))
+                    .build();
+            // @formatter:on
+            aclView.setAcl(Collections.singletonList(entry));
```
tonygermano: I think the local admin group should also be added with full control?

root always has access in linux, but windows does not automatically give permission to local admins.

server/conf/mirth.properties
```diff
@@ -25,10 +25,13 @@ password.reuselimit = 0
 version = 4.6.0
 
 # keystore
+# The keystore is created on first boot with no passphrase and is instead protected by file
+# permissions that restrict it to the user the server runs as. Set storepass/keypass if you would
+# rather protect it with a passphrase, and set type to JCEKS for keystores created before 4.6.0.
 keystore.path = ${dir.appdata}/keystore.jks
```
tonygermano: Should the file extension change to reflect the new file type?

I don't know when I'll get time to test this myself, but does the channel level encryption still work when the keystore type is PKCS12?

The second use of this keystore is to store the symmetric key for channel encryption. I suspect that's why the default type had not been previously changed, but maybe it's fine.

This unresolved issue and comments may be relevant https://github.com/nextgenhealthcare/connect/issues/4158

tonygermano (changes requested):

server/src/main/java/com/mirth/connect/server/util/FilePermissionUtil.java
mgaffigan: I agree in principle, but it is a pain.  Added as a separate commit in case we change our minds.

server/conf/mirth.properties
mgaffigan: > Should the file extension change to reflect the new file type?

Sure.  Updated.  I don't see a problem with this.

> does the channel level encryption still work when the keystore type is PKCS12?

If channel encryption is the "Encrypt message contents" checkbox, then yes, it seems to work fine.  Database rows seem to reference some AES key:

```text
{alg=AES/CBC/PKCS5Padding,cs=UTF-8,iv=ubupD9l5A2edZnhMyrE63Q==}zotLK9oahMRMDgLFoU4NHECvDfrRSrpwDD4YAjj6W9AE/89CFjyQCOIgN5N/xJUqr4sUzQTMMwbZ
mYtM8fwrk6T32m30b30TLbsuPXQs6MdbHwBY4Coi/ttOxmc0ZHfR46YBgd1LbUzPQNgwoKGZAlfJ
kI61PV1f8vdIDXB9gD3wNw9VegZTSUrbbYoVGvD6RjZdMHGuKqEBfrZaWqD1VnjV63Ii7TKy3GtR
iyPTKbDkEebzj2zJL3H26/I3n6T8IEw058nyO+gs48fOeZKFfBcH7S/6FX2HmrD8pA38LsxDbL+8
+dHJst+N68NwhyYU
```

server/basedir-includes/configure-from-env
mgaffigan: I removed it in favor of a filename check.

server/src/test/resources/mirth.properties
mgaffigan: sorry.  I was off: this is a test resource, not a published asset.

PR #423, opened 2026-08-26
Update default password policy to match 2026 nist minimums

mgaffigan:
Related to #227 

Affects new installations only

### DIFF
--- a/server/conf/mirth.properties
+++ b/server/conf/mirth.properties
@@ -9,13 +9,13 @@ http.port = 8080
 https.port = 8443
 
 # password requirements
-password.minlength = 0
+password.minlength = 15
 password.minupper = 0
 password.minlower = 0
 password.minnumeric = 0
 password.minspecial = 0
-password.retrylimit = 0
-password.lockoutperiod = 0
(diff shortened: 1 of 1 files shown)


server/conf/mirth.properties
```diff
@@ -9,13 +9,13 @@ http.port = 8080
 https.port = 8443
 
 # password requirements
-password.minlength = 0
+password.minlength = 8
```
gibson9583: Shouldn't this be 15?

"Verifiers and CSPs SHALL require passwords that are used as a single-factor authentication mechanism to be a minimum of 15 characters in length. Verifiers and CSPs MAY allow passwords that are only used as part of multi-factor authentication processes to be shorter but SHALL require them to be a minimum of eight characters in length."

https://pages.nist.gov/800-63-4/sp800-63b/authenticators/

mgaffigan: Updated.

gibson9583 (approved):

jonbartels (approved):

tonygermano (approved):

PR #427, opened 2026-08-29
Check password requirements on login

mgaffigan:
Changing password requirements does not affect current users. This PR modifies the login process to check requirements and handle them the same as password expiration, using the existing grace period functionality.

image

Note: `command` does not seem to handle expired passwords well.

Note to reviewers: Most of this change-set is whitespace changes due to indentation. See https://github.com/OpenIntegrationEngine/engine/pull/427/changes?w=1

### DIFF
--- a/client/src/main/java/com/mirth/connect/client/ui/ChangePasswordDialog.java
+++ b/client/src/main/java/com/mirth/connect/client/ui/ChangePasswordDialog.java
@@ -55,6 +55,12 @@ public class ChangePasswordDialog extends MirthDialog {
         });
 
         pack();
+
+        // Message can be several lines, grow to fit
+        passwordTextArea.setSize(passwordTextArea.getWidth(), Short.MAX_VALUE);
+        passwordPane.setPreferredSize(new Dimension(passwordPane.getWidth(), passwordTextArea.getPreferredSize().height));
+        pack();
+
         Dimension dlgSize = getPreferredSize();
         Dimension frmSize = parent.getSize();
         Point loc = parent.getLocation();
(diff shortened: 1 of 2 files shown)


NicoPiel (approved):

pacmano1:
What does this do when I try to grab a token to make API calls?  If it denies the token, we need to work through that.

mgaffigan:
@pacmano1 
> What does this do when I try to grab a token to make API calls? If it denies the token, we need to work through that.

Nothing in the above limits logins, it just returns the flag that the password needs to be changed in the response to the login.  Exactly the same as an "expired" password.

gibson9583:
> @pacmano1
> 
> > What does this do when I try to grab a token to make API calls? If it denies the token, we need to work through that.
> 
> Nothing in the above limits logins, it just returns the flag that the password needs to be changed in the response to the login. Exactly the same as an "expired" password.

@mgaffigan  are you sure? wont this affect a few things:

1. CLI - anything that isnt a plain `SUCCESS` gets a cant log into server.
2. grace period - with a >0 grace period or a -1 grace period, it can cause lockouts (but that may be OK, but contradicts your statement)

mgaffigan:
@gibson9583 
> @mgaffigan  are you sure? wont this affect a few things:

I'm sure that it uses the exact same response and grace period as an expired password. No new response statuses were added, nor any new grace period logic. 

> 1. CLI - anything that isnt a plain `SUCCESS` gets a cant log into server.

Agreed, that seems like a bug in CLI. It is the current expired password behavior, though. 

> 2. grace period - with a >0 grace period or a -1 grace period, it can cause lockouts

Agreed, but those have to be manually configured, and is the existing behavior of expired passwords.

PR #431, opened 2026-09-02
fix(dicom): EXPECT_FRAG crash on undefined-length UN sequences in fromXML

### LINKED ISSUE #430: [BUG] DICOM channels with a filter/transformer fail with `IllegalStateException: state:EXPECT_FRAG` on private undefined-length sequences
## Describe the bug

Any DICOM-typed connector that has a filter or transformer attached fails to process messages containing an undefined-length sequence under a private tag that dcm4che2 resolves to VR `UN` (implicit VR, tag not in its dictionary). Common real-world producers: FujiFILM `(0029,E131)` and Siemens MEDCOM `(0029,1140)`. Messages with no filter/transformer pass through fine, because they never go through the XML round trip.

```
com.mirth.connect.donkey.model.message.MessageSerializerException: Error converting XML to DICOM
Caused by: java.lang.IllegalStateException: state:EXPECT_FRAG
        at org.dcm4che2.io.ContentHandlerAdapter.startElement(...)
        at com.mirth.connect.plugins.datatypes.dicom.DICOMSerializer.fromXML(...)
```

## Root cause

dcm4che2 2.0.29's XML round trip is asymmetric:

- `DicomInputStream` reads a `UN` element with undefined length whose items hold dataset elements and represents it as `SQ` in memory.
- `toXML` uses the streaming `SAXWriter`, which emits the wire VR: `<tag0029E131 vr="UN" len="-1">` with structured `<item>` children.
- `fromXML` hands that to `ContentHandlerAdapter`, which only enters its sequence state for `vr="SQ"` and otherwise expects fragments; the first `<attr>` inside the item throws `state:EXPECT_FRAG`.

dcm4che2 has had no release since 2015, so this cannot be fixed at the library.

## To Reproduce

Programmatic, no fixtures needed (this is what the accompanying test does):

1. Build a `BasicDicomObject`, `putNestedDicomObject(0x0029E131, item)` with any element in the item, write it as Implicit VR Little Endian.
2. `new DICOMSerializer().toXML(base64)` — note the output carries `vr="UN"` with an `<item>` child.
3. `fromXML(thatXml)` throws.

Or in a running engine: DICOM Listener → any transformer step → send a Fuji or Siemens CR/CT object carrying a private sequence.

## Expected behavior

`fromXML` accepts the XML `toXML` produced, and the re-encoded DICOM is identical to what the engine emits without a transformer.

## Proposed fix

Reader-side, in `DICOMSerializer.fromXML`'s existing DOM pre-pass (the one that renames `<tagXXXXXXXX>` back to `<attr>`): when an element's `<item>` children contain renamed `<attr>` elements, set `vr="SQ"` before the SAX parse. That is the same content-based decision `DicomInputStream` makes for the binary form. `toXML` output is unchanged, so transformer scripts see identical XML; fragment containers (encapsulated pixel data) are untouched; every input that parses today produces an identical `DicomObject`; and the rewrite is a no-op if a future dcm4che emits `SQ` itself.

Verified byte-for-byte that the transformer round trip now produces the same DICOM as the no-transformer pass-through path, for implicit and explicit VR LE. PR to follow.

## Environment

- OIE `main` at 423df182f (also reproduces on 4.6.0-based builds)
- dcm4che2 2.0.29 as bundled

jessesaga:
Fixes #430

## Summary

DICOM channels with a filter or transformer fail with `IllegalStateException: state:EXPECT_FRAG` on messages that carry an undefined-length sequence under a private tag dcm4che2 resolves to VR `UN` (FujiFILM `0029,E131`, Siemens MEDCOM `0029,1140`, and similar). Root cause is an asymmetry inside dcm4che2 2.0.29's XML round trip: `SAXWriter` emits the wire VR (`vr="UN"`) for such sequences, but `ContentHandlerAdapter` only recognises `vr="SQ"` as a sequence and treats everything else as fragments.

This PR fixes it reader-side in `DICOMSerializer.fromXML`, inside the DOM pre-pass that already exists there: an element whose `<item>` children contain dataset elements gets `vr="SQ"` before the SAX parse. 25 lines of code plus a helper.

## Why this shape

- **Same decision the binary reader makes.** `DicomInputStream` decides SQ-vs-fragments by item content, not by VR. The predicate here ("item has an `<attr>` child") mirrors that, so the resulting `DicomObject` matches what a direct binary parse produces. I checked the alternative of keying on `vr="UN" len="-1"` and it diverges from the binary reader on a sequence whose items are all empty, so I did not use it.
- **No change to what transformer scripts see.** `toXML` is untouched. Fixing the writer to emit `SQ` would silently change `@vr` for existing scripts.
- **No change to emitted DICOM.** The bytes coming out of the transformer path are byte-identical to the no-transformer pass-through path, for implicit and explicit VR Little Endian.
- **Strictly additive.** Every XML that parses today produces the identical object; only inputs that threw before change behaviour. Encapsulated pixel data (fragment items holding only text) never trips the predicate.
- **Self-retiring.** The rewrite is skipped when `vr` is already `SQ`, so it becomes a no-op if the DICOM backend ever emits `SQ` itself.

## Testing

- New `DICOMSerializerUnSequenceTest` (JUnit 4, builds its DICOM programmatically, no binary fixtures):
  - `testUndefinedLengthUnSequenceRoundTrip`: asserts the `toXML` output actually carries the broken wire shape (`vr="UN"` with a structured `<item>`), then round-trips and checks the nested element and a defined-length `UN` blob survive.
  - `testEncapsulatedPixelDataFragmentsUntouched`: regression guard for the fragment path.
- On unpatched `main`, the first test fails with the exact production signature (`state:EXPECT_FRAG`). Patched, `./gradlew test -PdisableSigning=true` is green.

Note on naming: the existing `DICOMSerializerTests` is not picked up by the `**/*Test.class` filter (and references `tests/test-dicom-*.dcm` fixtures that are not in the repo), so the new class follows the `*Test` convention the build actually runs.

### DIFF
--- a/server/src/main/java/com/mirth/connect/plugins/datatypes/dicom/DICOMSerializer.java
+++ b/server/src/main/java/com/mirth/connect/plugins/datatypes/dicom/DICOMSerializer.java
@@ -122,6 +122,31 @@ public class DICOMSerializer implements IMessageSerializer {
                 }
             }
 
+            /*
+             * dcm4che2's XML round trip is asymmetric for sequences read from implicit-VR private
+             * tags: SAXWriter emits the wire VR (vr="UN" len="-1") with structured <item>
+             * children, but ContentHandlerAdapter only enters its sequence state for vr="SQ" and
+             * otherwise expects fragments, throwing IllegalStateException("state:EXPECT_FRAG")
+             * when a dataset item arrives. DicomInputStream already treats such elements as SQ in
+             * memory, so present them to the parser the same way. Only items holding renamed
+             * <attr> children mark a dataset sequence; fragment containers (encapsulated pixel
+             * data, whose items hold only base64/hex text) are left untouched.
+             */
+            if (items != null) {
+                for (int i = 0; i < items.getLength(); i++) {
+                    Node itemNode = items.item(i);
+                    Node parentNode = itemNode.getParentNode();
+
+                    if (parentNode instanceof Element && hasAttrElementChild(itemNode)) {
+                        Element parentElement = (Element) parentNode;
+
+                        if (parentElement.hasAttribute("vr") && !parentElement.getAttribute("vr").equals("SQ")) {
+                            parentElement.setAttribute("vr", "SQ");
+                        }
+                    }
+                }
+            }
+
             // find the charset
             String charset = null;
             Element charsetElement = (Element) document.getElementsByTagName("tag00080005").item(0);
@@ -207,6 +232,16 @@ public class DICOMSerializer implements IMessageSerializer {
                 document.renameNode(node, null, "tag" + tag);
             }
         }
+    }
+
+    private boolean hasAttrElementChild(Node node) {
+        for (Node child = node.getFirstChild(); child != null; child = child.getNextSibling()) {
+            if (child.getNodeType() == Node.ELEMENT_NODE && child.getNodeName().equals("attr")) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private void renameTagToAttr(Document document, Node node) throws DOMException {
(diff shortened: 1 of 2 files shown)


mgaffigan (commented):
The code looks reasonable, but I'd like to hear from someone who knows DICOM before approving.  

From a cleanliness perspective, I don't love that we're editing the XML after construction rather than fixing the generator, though.

jessesaga:
Converting back to draft. Digging into this further based on the suggestions turned up a few issues I need to address.

PR #439, opened 2026-09-10
Stop failed logins from revealing which usernames exist

### LINKED ISSUE #240: [BUG] Username enumeration possible if lockout enabled

pacmano1:
Fixes #240.

A failed login gave away whether the username was real, in two separate ways. This fixes both.

### The messages

The lockout paths in `DefaultUserController.authorizeUser` named the account and reported how long the lock had left to run, so an unauthenticated caller could tell a valid username from an invalid one. Both sites now return a generic `Incorrect username or password.` unless `password.allowdetailedautherrors` is set, which defaults to false.

### The timing

Fixing the messages alone left a second and much larger signal in place. Closing #240 on the message fix by itself would have overstated what changed.

The password digest is only reached for a valid, unlocked user. An unknown username never enters the `validUser != null` block, and a locked account returns before it, so both skipped 600,000 PBKDF2 iterations. Measured with generic messages already enabled, 250 samples per cohort:

| cohort | median | p10 | p90 |
|---|---|---|---|
| valid + unlocked | 230.90 ms | 227.52 | 234.27 |
| valid + locked | 0.49 ms | 0.42 | 0.72 |
| invalid username | 0.76 ms | 0.66 | 0.93 |

The distributions do not overlap. The fastest unlocked sample was still 240x slower than the slowest locked one, so one timed request classified any username whatever the response body said. An attacker could enumerate accounts without reading a message at all.

A same-account control confirms the cause. An account that crossed the strike limit partway through a run stepped from 232.03 ms to 0.33 ms at the boundary, with no samples in between. The account did not change, only its lockout state did.

Every path that skips the real digest now runs a throwaway one and discards the result: an unknown username, a locked account, an account that exists but has never had a password set, and a failed pre-2.2 hash check. The last two are reachable in normal use, since an admin can create an account before provisioning it and legacy hashes survive upgrades. The throwaway goes through `matches()` so the fallback digester runs the way it does on the real path, and the hash is generated once from the live digester so it follows the configured algorithm and iteration count. Measured again on the same rig by the same method:

| cohort | before | after |
|---|---|---|
| valid + unlocked | 230.90 ms | 237.74 ms |
| valid + locked | 0.49 ms | 236.84 ms |
| invalid username | 0.76 ms | 237.05 ms |
| worst ratio | 470x | 1.00x |

The no-password case was checked separately against a real account and came back at 234.16 ms against 233.89 for a valid user and 233.45 for an unknown one.

This is not constant time and I am not claiming it is. A valid user still costs one extra credentials query that a miss does not, which measured under a millisecond against a 237 ms floor. There is a `ponytail:` marker on the helper naming that ceiling and the condition for revisiting it.

There is a tradeoff here worth arguing with. Locked accounts are no longer cheap to reject, so lockout no longer caps the cost of repeatedly probing an account already known to exist. My reasoning is that an attacker gains nothing by locking an account once every username costs the same, so that cheapness was mostly helping the attacker. Tell me if you see it differently.

### What this does not cover

An `AuthorizationPlugin` returning a non-null `LoginStatus` returns before any of this code, so OIDC, LDAP and similar plugins are unaffected in both directions: nothing here breaks them, and nothing here protects them. Their messages and timing are their own. A plugin that replaces `UserController` outright through `ExtensionLoader` would miss this entirely, and grepping this repo will not reveal one.

### Tests

`DefaultUserControllerAuthorizeUserTest` covers the message behaviour: a locked account is generic by default, reports detail when the property is set, and is indistinguishable from an unknown username. The first and third fail against current main and pass with this change, so they are regression tests rather than a description of the new code. The second passes either way by design, since the detailed message used to be unconditional; it guards the escape hatch.

It binds mock `ControllerFactory` and `SqlConfig` through Guice static injection, the way `DefaultHttpConfigurationTest` already does, so no database is involved. Lockout state is set directly on the `User`, because `LoginRequirementsChecker` reads the strike count and last strike time straight off it.

No test asserts the timing. Mocking the digester removes the cost being measured, and a real one makes every call take 230 ms and turns the suite into a flaky performance test. The numbers above come from a running server.

**Not added: a smoke test in `ci/`.** The new harness boots a real engine across seven configurations and has a Java seam with a live `Client`, so the message behaviour could be asserted there against all five databases. That is worth doing and I have not done it; it belongs in its own change rather than growing this one further. The timing assertion would still be a poor fit for shared runners.

Clean build, 686 tests, 0 failures.

### Credit

This is @mgaffigan's fix from #241, reworked for the current source layout. That branch predates the June move to `src/main/java` and no longer applies. His PR should close in favour of this one. The timing half is new.

### Naming

Uses `allowDetailedAuthErrors`, which @tonygermano suggested on #241 and @mgaffigan declined, wanting a name that sounds alarming rather than helpful.

I have gone with Tony's, on a ground that was not available in January. The measurements above show that suppressing the messages does not by itself prevent username enumeration. A flag called `allowUsernameEnumeration` would promise a guarantee the flag cannot keep, and an admin who set it false would reasonably believe the hole was closed. `allowDetailedAuthErrors` describes exactly what it controls, which after this PR is only the messages, because enumeration is now handled separately and unconditionally.

### Constructor

The field is set explicitly in both constructors, per @tonygermano's note that every other property is. I have not added a twelfth constructor argument. Production builds this through the no-arg constructor and setters, so widening the signature would only churn `SwaggerExamplesServlet` and the existing tests to thread a parameter nothing reads.

### Test rename

`PasswordRequirementsTests` becomes `PasswordRequirementsTest`. The build includes `**/*Test.class` only, so the plural-named class has never run in any build. Six dormant cases now execute and pass, alongside two new ones for the property. The dead `*Tests` problem is wider than this file and needs its own issue.

### Behavior change

With the property unset, messages go from detailed to generic. That is deliberate and secure by default, but it is still a change, and anything scraping the message string will notice. Raised on #241 by me and @ChristopherSchultz.

### How to verify by hand

Defaults are `retrylimit = 100` and `lockoutperiod = 1`. That period is in hours rather than minutes, per `Duration.standardHours` at `LoginRequirementsChecker:99`.

1. `POST /api/users/_login` with a valid username and wrong password, 101 times
2. Send the same request against a nonexistent username. The responses are byte-identical
3. Retry the valid user with the correct password. It still fails generically, which confirms the lock is real and the gated branch ran
4. Time all three cases. They land within a millisecond or two of each other instead of 300x apart
5. Set `password.allowdetailedautherrors = true`, restart and repeat. The account name and remaining time come back, and the timings stay equal

### DIFF
--- a/server/conf/mirth.properties
+++ b/server/conf/mirth.properties
@@ -21,6 +21,11 @@ password.graceperiod = 0
 password.reuseperiod = 0
 password.reuselimit = 0
 
+# When false (the default), failed logins return a generic "Incorrect username or password."
+# message. Set to true to restore detailed messages that reveal account lockout state and
+# remaining attempts. Warning: enabling this lets an attacker enumerate valid usernames.
+password.allowdetailedautherrors = false
+
 # Only used for migration purposes, do not modify
 version = 4.6.0
 
--- a/server/src/main/java/com/mirth/connect/model/PasswordRequirements.java
+++ b/server/src/main/java/com/mirth/connect/model/PasswordRequirements.java
@@ -29,6 +29,7 @@ public class PasswordRequirements implements Serializable {
     private int gracePeriod;
     private int reusePeriod;
     private int reuseLimit;
+    private boolean allowDetailedAuthErrors;
 
     public PasswordRequirements() {
         this.minLength = 0;
@@ -42,6 +43,7 @@ public class PasswordRequirements implements Serializable {
         this.gracePeriod = 0;
         this.reusePeriod = 0;
         this.reuseLimit = 0;
+        this.allowDetailedAuthErrors = false;
     }
 
     public PasswordRequirements(int minLength, int minUpper, int minLower, int minNumeric, int minSpecial, int retryLimit, int lockoutPeriod, int expiration, int gracePeriod, int reusePeriod, int reuseLimit) {
@@ -56,6 +58,7 @@ public class PasswordRequirements implements Serializable {
         this.gracePeriod = gracePeriod;
         this.reusePeriod = reusePeriod;
         this.reuseLimit = reuseLimit;
+        this.allowDetailedAuthErrors = false;
     }
 
     public int getMinLength() {
@@ -144,5 +147,13 @@ public class PasswordRequirements implements Serializable {
 
     public void setReuseLimit(int reuseLimit) {
         this.reuseLimit = reuseLimit;
+    }
+
+    public boolean getAllowDetailedAuthErrors() {
+        return allowDetailedAuthErrors;
+    }
+
+    public void setAllowDetailedAuthErrors(boolean allowDetailedAuthErrors) {
+        this.allowDetailedAuthErrors = allowDetailedAuthErrors;
     }
 }
--- a/server/src/main/java/com/mirth/connect/server/controllers/DefaultUserController.java
+++ b/server/src/main/java/com/mirth/connect/server/controllers/DefaultUserController.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -43,6 +44,12 @@ import com.mirth.connect.server.util.StatementLock;
 public class DefaultUserController extends UserController {
     public static final String VACUUM_LOCK_PERSON_STATEMENT_ID = "User.vacuumPersonTable";
     public static final String VACUUM_LOCK_PREFERENCES_STATEMENT_ID = "User.vacuumPersonPreferencesTable";
+    private static final String INCORRECT_CREDENTIALS_MESSAGE = "Incorrect username or password.";
+    private static final String TIMING_EQUALIZATION_PASSWORD = "password used only to equalize authentication timing";
+
+    private final Object timingEqualizationLock = new Object();
+    private volatile String timingEqualizationHash;
+    private final AtomicBoolean timingEqualizationFailureLogged = new AtomicBoolean();
 
     private Logger logger = LogManager.getLogger(this.getClass());
     private ExtensionController extensionController = null;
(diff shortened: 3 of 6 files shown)


mgaffigan (approved):
I think this is unlikely to prevent timing side channel analysis, but it's better than nothing.

pacmano1:
> I think this is unlikely to prevent timing side channel analysis, but it's better than nothing.

Far from perfect.   Not really sure how to best do this by piling on more stuff, especially given the variation in machines and configs.   Maybe change authroizeUser a lot or look into rate limiting, etc.   At some threshold the behavior does have a pattern.   And a single user name look up make the variation after a number of calls easier to see.

mgaffigan:
> > I think this is unlikely to prevent timing side channel analysis, but it's better than nothing.
> 
> Far from perfect. Not really sure how to best do this by piling on more stuff, especially given the variation in machines and configs. Maybe change authroizeUser a lot or look into rate limiting, etc. At some threshold the behavior does have a pattern. And a single user name look up make the variation after a number of calls easier to see.

Agreed.  When a secure implementation is required, my answer remains "Don't try and do user auth; it's deceptively hard.  Use an a separate OIDC IdP."

PR #442, opened 2026-09-15
feat: add optional message telemetry hooks for OpenTelemetry integration

### LINKED ISSUE #432: [IDEA] Add support for telemetry plugins: a message lifecycle listener SPI in donkey
**Is your feature request related to a problem? Please describe.**

Plugins cannot observe message processing while it happens. The engine exposes counters per channel and connector (received, filtered, sent, errored), queue sizes, and connection state events, but nothing tells a plugin that a specific message was received, transformed, queued, sent, or failed, on which thread, and when. The existing `MessageEvent` is a statistics delta: it carries no message id, drops the TRANSFORMED and PENDING transitions, and fires from `Statistics.update` at transaction commit rather than at the transition. Anyone who wants per-message telemetry today has to fork the engine, wrap connectors, or parse logs.

**Describe your use case**

I need this so that telemetry can be delivered as ordinary plugins instead of engine changes. Examples of plugins this enables, none of which the engine itself needs to know about:

- an OpenTelemetry or Prometheus exporter that emits per-message traces and latency histograms
- an audit or compliance hook that records status transitions for selected channels
- an SLA monitor that alerts when a message spends too long in a queue or a destination

The engine stays vendor neutral. The SPI is the only engine change; every exporter is a plugin.

gibson9583:
## Summary

Adds an optional message telemetry bridge to the engine so an extension can observe message processing and maintain trace context across destination-chain and JavaScript worker handoffs.

This provides the engine integration points needed by the OpenTelemetry plugin. SDK setup, exporters, sampling, configuration and durable trace-continuity policy remain extension responsibilities.

Closes #432  

## Changes

- Add a `MessageTelemetry.Provider` contract with observations for source processing, destination attempts, filtering/transformation, sending and response transformation.
- Transfer context across destination-chain and JavaScript executors, activating it only when work starts and restoring the worker's previous context when it finishes.
- Add an ingress callback before the source map becomes read-only and is first persisted, allowing a provider to prepare private propagation metadata.
- Carry optional dispatch context through native source-to-destination and destination-to-destination handoffs. The context uses a transient, owner-keyed slot on `ConnectorMessage`, with compare-and-set publication to prevent older or reentrant callbacks from overwriting a newer reservation.
- Keep destination observations open through retry handling and engine cleanup, including DAO/queue cleanup attempts and status-lock release.
- Document the provider contract, lifecycle boundaries and behavior matrix.

## Compatibility and failure handling

- No new runtime dependencies or database schema changes.
- With no provider installed, observation hooks are no-ops and executor wrapping returns the original task.
- Dispatch context is excluded from Java and XML serialization; the existing Java serialization identity is preserved. Providers must keep this context bounded, immutable and free of message content or live SDK resources.
- Ordinary provider callback failures are isolated from message processing and produce one fixed warning per installation, without exception content. VM errors and `ThreadDeath` retain fatal handling.
- Provider registration allows one active installation; closing an old registration cannot detach a newer provider.

## Validation

Regression coverage includes synchronous and queued processing, parallel destinations, retries, persisted PENDING messages, filtering and transformation failures, callback failures, cleanup ordering, executor rejection/cancellation, and JavaScript context restoration.


jonbartels:
@gibson9583 - I have not reviewed yet, but one of the first things that jumps out at me is that the PR needs links or docs for:
* wheres the plugin?
* Why can't the plugin add its own endpoints instead of modifying core OIE?
* what data elements are exposed?
* are there docker-compose files, integration tests, etc that would be useful for testing or demonstrating what this does?

PR #451, opened 2026-09-19
Fix SQLi in DatabaseConnectorServlet (CVE-2026-82583)

mgaffigan:
Prior code was attempting to retrieve 0-1 rows in a per-driver manner,
which is hard.  Switched to using the standard `WHERE 1 = 0` approach.

Left the now-pointless parameter in place for compatibility with existing callers.

Review notes:
* See the [Fix SQLi in DatabaseConnectorServlet](https://github.com/OpenIntegrationEngine/engine/commit/81241a14ecdbfea6068f48e481aa3f2c87fe9c24) commit for the actual fix
* Most of the PR is tests.  The functionality is now tested across all supported engines.
* Drive-by fix for improper quoting of table names


jonbartels (approved):

pacmano1:
Nothing exercises the servlet with a non-null `selectLimit`, so deleting the `selectLimit = null;` line puts the CVE back with every test still green.

Can you add a regression test? #452 and #453 each carry one, this is the only one of the three without.

The fix works. I verified it against the smoke harness on Derby: a caller-supplied `selectLimit` that aliases a column no longer reaches `executeQuery`, and with that line removed as a control, the same check fails and returns the injected alias. Aliasing a column and checking whether the alias comes back in the metadata works at whatever layer you prefer to test.

mgaffigan:
> Nothing exercises the servlet with a non-null `selectLimit`, so deleting the `selectLimit = null;` line puts the CVE back with every test still green.

The selectLimit parameter is not referenced - so there is nothing to exercise.  The `selectLimit = null` is just for clarity and to avoid a mental phi node.

> Can you add a regression test? #452 and #453 each carry one, this is the only one of the three without.

It's challenging to write any meaningful test for this - the unit tests do not run with a database, and since the parameter is not referenced it is a bit pointless to assert that it is not present in the generated SQL.

The correct fix is the absence of the code.  We don't have to test the code since it does not exist.  We test the feature, though, but we can't meaningfully test for the absence of user input that might end up in the query.  (To extreme, we can remove the parameter, which would guarantee that it is unused, but cause minimal compat issues).

server/src/main/java/com/mirth/connect/connectors/jdbc/DatabaseConnectorServlet.java
```diff
@@ -52,15 +41,16 @@ public DatabaseConnectorServlet(@Context HttpServletRequest request, @Context Se
 
     @Override
     public SortedSet<Table> getTables(String channelId, String channelName, String driver, String url, String username, String password, Set<String> tableNamePatterns, String selectLimit, Set<String> resourceIds) {
+        // selectLimit is deprecated and ignored for security reasons. Kept for backcompat.
+        selectLimit = null;
```
tonygermano, line 45: Following up on @pacmano1's question, can we deprecate this entire method and have it call an overloaded method that drops the selectLimit parameter?

server/src/main/java/com/mirth/connect/connectors/jdbc/TableMetadataReader.java
```diff
@@ -0,0 +1,219 @@
...
+    static String buildColumnQuery(String identifierQuote, String schema, String tableName) {
+        // The JDBC contract returns a single space when the database does not support quoting.
+        final String quote = StringUtils.trimToNull(identifierQuote);
+        if (quote == null) {
+            return null;
+        }
+
+        final String quotedTableName = quote(quote, tableName);
+        final String schemaTableName = StringUtils.isNotEmpty(schema) ? quote(quote, schema) + "." + quotedTableName : quotedTableName;
+        return "SELECT * FROM " + schemaTableName + " WHERE 1 = 0";
```
tonygermano, line 161: praise: I agree with always returning 0 rows instead of 0-1 rows and removing the need for the select limit.

gibson9583 (approved):

jonbartels (approved):

server/src/main/java/com/mirth/connect/connectors/jdbc/DatabaseConnectorServlet.java
mgaffigan: @pacmano1 / @tonygermano,

I can, I have a commit hanging around somewhere that does this, but it increases things substantially.  I am happy to put a separate PR that strips the functionality in the interest of getting this merged.  If you think this is blocking, lmk.

pacmano1: not blocking for me.

mgaffigan: Ok.  I'll open after this closes.  Only one more review needed.

jonbartels (approved):

server/src/main/java/com/mirth/connect/connectors/jdbc/DatabaseConnectorServlet.java
mgaffigan: @tonygermano, do you need this in this PR? We're good to merge, otherwise. I'll open the "clean up" as soon as this merges, but that's not urgent to merge.

tonygermano: I'm fine with handling in a follow-up

tonygermano (approved):

server/src/main/java/com/mirth/connect/connectors/jdbc/DatabaseConnectorServlet.java
jonbartels: https://github.com/OpenIntegrationEngine/engine/issues/461 to capture this issue for followup.

PR #452, opened 2026-09-19
Fix XXE in XML batch parsing

mgaffigan:
CVE-2026-82578 found an XXE vulnerability in XML batch parsing. This closes that vulnerability and adds regression tests.

Review notes:
* See [Fix XXE in XML batch parsing](https://github.com/OpenIntegrationEngine/engine/commit/f1d73696bb89b5ebbb270e393dc6873c5f544a30) for the actual fix
* Most of this is tests, since the feature was not previously tested


jonbartels (dismissed):

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
```diff
@@ -133,6 +134,32 @@ public Long processMessage(final String channelId, final RawMessage rawMessage)
...
+
+        try {
+            engineController.dispatchRawMessage(channelId, rawMessage, true, true, responseHandler);
+
+            List<Long> messageIds = new ArrayList<Long>();
+            for (DispatchResult dispatchResult : responseHandler.getDispatchResults()) {
+                messageIds.add(dispatchResult.getMessageId());
+            }
+
+            containerRequestContext.setProperty(ResponseCodeFilter.RESPONSE_CODE_PROPERTY, Response.Status.CREATED.getStatusCode());
```
gibson9583, line 145: When a channel is stopping, or its batch factory has stopped accepting work, dispatch can return without accepting any messages. This method still returns `201 Created` with an empty list. A caller can treat that as success and never retry the dropped submission.

I reproduced both cases. The existing message endpoint reports failure in the same situation. Please distinguish a refused batch from a valid empty batch, report the refusal as a failure, and add tests for both shutdown cases.

donkey/src/main/java/com/mirth/connect/donkey/server/message/batch/CollectingResponseHandler.java
```diff
@@ -0,0 +1,32 @@
...
+public class CollectingResponseHandler extends SimpleResponseHandler {
+
+    private final List<DispatchResult> dispatchResults = new ArrayList<DispatchResult>();
+
+    @Override
+    public void setDispatchResult(DispatchResult dispatchResult) {
+        super.setDispatchResult(dispatchResult);
+
+        if (dispatchResult != null) {
+            dispatchResults.add(dispatchResult);
```
gibson9583: Please collect only the message IDs needed by the response. Keeping every `DispatchResult` also keeps processed messages, maps, and content in memory until the whole batch finishes. Large batches can run out of memory even after earlier messages finish processing.

In an isolated comparison, the new collector ran out of memory while the existing handler completed the same workload.

gibson9583 (dismissed):
Requesting changes before merge.

The new batch endpoint can report success without accepting a batch, and its response collector retains all processed message payloads. Please address the two inline comments.

The Oracle integration check is also failing in the new CSV batch test: `Expected source transformed content but the server stored none`. Please investigate and get a passing Oracle run before merging. I have not established whether the missing content is caused by this PR.

[Failing Oracle check](https://github.com/OpenIntegrationEngine/engine/actions/runs/35471826818/job/105974735625)

donkey/src/main/java/com/mirth/connect/donkey/server/message/batch/CollectingResponseHandler.java
mgaffigan: Accepted.

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
mgaffigan: Agreed that this is an issue - but it seems to be pre-existing and general.  Tracked in https://github.com/OpenIntegrationEngine/engine/issues/459

mgaffigan:
@gibson9583, the oracle issue was opened with issue #454 and fixed in #455

gibson9583 (dismissed):
Items raised are non-blocking. Approved

tonygermano (changes requested):
My requests are for documentation improvements and not to challenge the code.

Can you expand the commit messages? Some of these changes are not small, and it would be good to have a summary of the changes and justification for why they are needed, especially when they touch multiple files or appear at a glance to change workflows.

Normally batch processing is controlled by the channel definition. What is the expected behavior when calling the new API method batchMessagesWithObj on a channel which doesn't have batch processing enabled? Is the only difference between calling this method and the existing method the return value? I think the method description implies that might be the case, but it should also describe what happens when called on a channel with batch processing disabled.

mgaffigan:
> Can you expand the commit messages? Some of these changes are not small, and it would be good to have a summary of the changes and justification for why they are needed, especially when they touch multiple files or appear at a glance to change workflows.

You should have write access to the branch.  Feel free to reword.  The commits are thematically made, already, so there should not be much explanation I can think of - and no issue is open to link to (nor do I think these are nuanced enough to warrant one).
 
> Normally batch processing is controlled by the channel definition. What is the expected behavior when calling the new API method batchMessagesWithObj on a channel which doesn't have batch processing enabled? Is the only difference between calling this method and the existing method the return value? I think the method description implies that might be the case, but it should also describe what happens when called on a channel with batch processing disabled.

Not sure - but the change is to _return_ the message ID's from process batch.  It does not change the processing.  Whatever happened previously would still happen, just now you have an endpoint which gives all of the message ID's instead of just 1.

pacmano1:
This rejects batches whose DOCTYPE declares only internal entities and points nowhere. I added a case to `210-xml-batch-xxe`: a batch declaring `<!ENTITY site "CLINIC-A">` with two messages. On main it splits into the two expected messages with the entity resolved; on this branch the server refuses it with a 500. `02-two-messages` passes on both, so the difference is the DOCTYPE rather than the batch path.

jonbartels flagged the same behaviour change on #441 on 15 Sep and offered a changelog entry, and abhinavagarwal07 asked there about a release note. Neither got a reply, and nothing here mentions it.

#453 goes the other way, permitting the DOCTYPE and setting only `ACCESS_EXTERNAL_DTD`. Measured alone, that setting errors on the external-entity payload and still accepts an internal-only DOCTYPE, so both approaches close the CVE and only this one drops traffic that works today. The two PRs should probably agree.

mgaffigan:
> This rejects batches whose DOCTYPE declares only internal entities and points nowhere. I added a case to `210-xml-batch-xxe`: a batch declaring `<!ENTITY site "CLINIC-A">` with two messages. On main it splits into the two expected messages with the entity resolved; on this branch the server refuses it with a 500. `02-two-messages` passes on both, so the difference is the DOCTYPE rather than the batch path.
> 
> jonbartels flagged the same behaviour change on #441 on 15 Sep and offered a changelog entry, and abhinavagarwal07 asked there about a release note. Neither got a reply, and nothing here mentions it.
> 
> #453 goes the other way, permitting the DOCTYPE and setting only `ACCESS_EXTERNAL_DTD`. Measured alone, that setting errors on the external-entity payload and still accepts an internal-only DOCTYPE, so both approaches close the CVE and only this one drops traffic that works today. The two PRs should probably agree.
> 

You and Tony are right. The rest of the app blocks DTDs entirely. I had tried to leave a small hole in the XSLT on the basis that someone might be using XHTML source messages, but it did not work, and I see no reasonable way to make it work.

Updated that PR to block DTD entirely, making it consistent with this PR (and the majority of the engine, which already blocks DTDs).

server/src/main/java/com/mirth/connect/plugins/datatypes/xml/XMLBatchAdaptor.java
```diff
@@ -127,7 +130,12 @@ private String getMessageFromReader() throws Exception {
 
                 XPath xpath = xPathFactory.newXPath();
 
-                nodeList = (NodeList) xpath.evaluate(query.toString(), new InputSource(bufferedReader), XPathConstants.NODESET);
+                // Parse the XML securely to prevent XXE
+                DocumentBuilderFactory documentBuilderFactory = DocumentSerializer.getSecureDocumentBuilderFactory();
+                documentBuilderFactory.setNamespaceAware(true);
```
tonygermano: question: is setting namespaceAware a behavior change?

If so, is it necessary? None of the tests use namespaces.

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
```diff
@@ -133,6 +134,27 @@ public Long processMessage(final String channelId, final RawMessage rawMessage)
...
+
+        try {
+            engineController.dispatchRawMessage(channelId, rawMessage, true, true, responseHandler);
+
+            containerRequestContext.setProperty(ResponseCodeFilter.RESPONSE_CODE_PROPERTY, Response.Status.CREATED.getStatusCode());
+            return responseHandler.getMessageIds();
+        } catch (ChannelException e) {
+            // Do nothing. An error should have been logged.
+        } catch (BatchMessageException e) {
+            logger.error("Error processing batch message", e);
```
tonygermano:
```suggestion
            logger.error("Error processing batch message for channel " + channelId, e);
```

tonygermano (changes requested):
See inline suggestion and question. I was going to have clanker write the missing commit messages bodies, but I did not know if they would be retained if other changes go through.

server/src/main/java/com/mirth/connect/plugins/datatypes/xml/XMLBatchAdaptor.java
mgaffigan: It is not a behavior change, no.  The prior xpath.evaluate did it by default, but it is not a default of `DocumentBuilderFactory`.  Added a regression test for proof.

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
mgaffigan: Accepted.

gibson9583 (approved):

gibson9583 (approved):

pacmano1 (approved):
The DOCTYPE question is settled, but the behavior change still needs a release note. A batch whose DOCTYPE declares only internal entities splits on 4.5 and gets a 500 after this PR. Can the PR body call that out so it lands in the 4.6.x notes? Same ask jonbartels made on #441.

PR #453, opened 2026-09-19
Harden XSLT step against XXE attacks

mgaffigan:
Parses the untrusted XML securely to prevent XXE attacks.  Retains trust of the XML Stylesheet since that is not attacker controlled.

Review notes:
* See [Harden XSLT step against XXE attacks](https://github.com/OpenIntegrationEngine/engine/commit/b1a5281ca50f37e82e8abcd7a330a98b970c9f89) for actual fix
* Most of this is regression tests


jonbartels (dismissed):

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
```diff
@@ -133,6 +134,32 @@ public Long processMessage(final String channelId, final RawMessage rawMessage)
...
+
+        try {
+            engineController.dispatchRawMessage(channelId, rawMessage, true, true, responseHandler);
+
+            List<Long> messageIds = new ArrayList<Long>();
+            for (DispatchResult dispatchResult : responseHandler.getDispatchResults()) {
+                messageIds.add(dispatchResult.getMessageId());
+            }
+
+            containerRequestContext.setProperty(ResponseCodeFilter.RESPONSE_CODE_PROPERTY, Response.Status.CREATED.getStatusCode());
```
gibson9583: When a channel is stopping, or its batch factory has stopped accepting work, dispatch can return without accepting any messages. This method still returns `201 Created` with an empty list. A caller can treat that as success and never retry the dropped submission.

I reproduced both cases. The existing message endpoint reports failure in the same situation. Please distinguish a refused batch from a valid empty batch, report the refusal as a failure, and add tests for both shutdown cases.

donkey/src/main/java/com/mirth/connect/donkey/server/message/batch/CollectingResponseHandler.java
```diff
@@ -0,0 +1,32 @@
...
+public class CollectingResponseHandler extends SimpleResponseHandler {
+
+    private final List<DispatchResult> dispatchResults = new ArrayList<DispatchResult>();
+
+    @Override
+    public void setDispatchResult(DispatchResult dispatchResult) {
+        super.setDispatchResult(dispatchResult);
+
+        if (dispatchResult != null) {
+            dispatchResults.add(dispatchResult);
```
gibson9583: Please collect only the message IDs needed by the response. Keeping every `DispatchResult` also keeps processed messages, maps, and content in memory until the whole batch finishes. Large batches can run out of memory even after earlier messages finish processing.

In an isolated comparison, the new collector ran out of memory while the existing handler completed the same workload.

server/src/main/java/com/mirth/connect/util/MirthXmlUtil.java
```diff
@@ -114,6 +119,18 @@ public static String prettyPrint(String input) {
...
     }
 
+    /** Returns a {@link Source} for XML from an untrusted origin. */
+    public static Source getSecureSource(Reader reader) throws Exception {
+        SAXParserFactory factory = SAXParserFactory.newInstance();
+        // False by default
+        factory.setNamespaceAware(true);
+
+        SAXParser parser = factory.newSAXParser();
+        parser.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
gibson9583: Adding Xerces 2.12.2 to a channel's resource library makes valid XSLT steps fail here. `SAXParserFactory.newInstance()` selects Xerces from the channel's classloader, but Xerces rejects the `ACCESS_EXTERNAL_DTD` property with `SAXNotRecognizedException`.

I reproduced this through the generated transformer script: a simple transformation of `<r>hello</r>` succeeds before this change and fails afterward. Both normal and iterator steps are affected.

Please use a parser that supports the required security settings, or configure other providers securely, and add a regression test with Xerces. Ignoring the property error would leave the security fix incomplete.

gibson9583 (dismissed):
Requesting changes before merge.

The new parsing path breaks valid XSLT transformations when a channel uses Xerces. The batch endpoint added in this PR also has two issues: it can report success without accepting a batch, and it retains all processed message payloads.

Please address the three inline comments. The two batch API issues also apply to PR #452 because both PRs include the same API change.

server/src/main/java/com/mirth/connect/server/api/servlets/MessageServlet.java
mgaffigan: Agreed that this is an issue - but it seems to be pre-existing and general.  Tracked in https://github.com/OpenIntegrationEngine/engine/issues/459

donkey/src/main/java/com/mirth/connect/donkey/server/message/batch/CollectingResponseHandler.java
mgaffigan: Accepted.

gibson9583 (dismissed):
good to merge, items raised are non-blocking

server/src/main/java/com/mirth/connect/util/MirthXmlUtil.java
tonygermano: I believe `SAXParserFactory factory = SAXParserFactory.newDefaultInstance();` would resolve this.

tonygermano (changes requested):
I had requested in #452 that the commit messages be expanded. That applies here, too, especially since this PR has the same or similar commit that prompted the request in the other PR.

mgaffigan:
@tonygermano
> I had requested in #452 that the commit messages be expanded. That applies here, too, especially since this PR has the same or similar commit that prompted the request in the other PR.

You're welcome to rewrite the commit messages - I said everything I would think to say in them (short of asking a clanker to bloviate).  If someone wants to know "what" changed, they have the git diff.  I'm focusing on the "why".

server/src/main/java/com/mirth/connect/util/MirthXmlUtil.java
mgaffigan: Updated.

gibson9583 (dismissed):

server/src/main/java/com/mirth/connect/util/MirthXmlUtil.java
```diff
@@ -114,6 +119,20 @@ public static String prettyPrint(String input) {
...
+        // Use newDefaultInstance to avoid whatever is on the classpath that might
+        // be poisoned from the channel classloader.
+        SAXParserFactory factory = SAXParserFactory.newDefaultInstance();
+        // False by default
+        factory.setNamespaceAware(true);
+
+        SAXParser parser = factory.newSAXParser();
+        parser.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+
+        return new SAXSource(parser.getXMLReader(), new InputSource(reader));
```
tonygermano: Suggestions based on OWASP recommendations when not using `disallow-doctype-decl`. These should all be supported since we are forcing the "default" (built-in) SAXParserFactory instance.

```suggestion
        // 1. Mitigate XML Bombs (Billion Laughs) limits
        factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);

        // 2. Defense-in-depth: Disable Xerces external entity and DTD features
        factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
        factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
        factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);

        SAXParser parser = factory.newSAXParser();
        XMLReader xmlReader = parser.getXMLReader();

        // 3. JAXP 1.5 protocol restriction: explicitly disable DTD access to override any system properties
        xmlReader.setProperty(XMLConstants.ACCESS_EXTERNAL_DTD, "");
        xmlReader.setProperty(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");

        return new SAXSource(xmlReader, new InputSource(reader));
```

tonygermano:
> @tonygermano
> 
> > I had requested in #452 that the commit messages be expanded. That applies here, too, especially since this PR has the same or similar commit that prompted the request in the other PR.
> 
> You're welcome to rewrite the commit messages - I said everything I would think to say in them (short of asking a clanker to bloviate). If someone wants to know "what" changed, they have the git diff. I'm focusing on the "why".

@mgaffigan No longer necessary on this PR. You removed the large duplicate commit that also appears in the other PR. The remaining changes are small enough to not need additional explanation.

server/src/main/java/com/mirth/connect/util/MirthXmlUtil.java
mgaffigan: Updated.  I had gone soft to try to allow XHTML to pass unmolested, but in testing, that did not work regardless.  matched the rest of the app with disabling DTD's entirely

jonbartels (approved):

tonygermano (approved):

gibson9583 (approved):

